PR #3, opened 2020-07-01
Samvera tulip tree

markpbaggett:
This completes my action item from the last meeting for creating mappings for the sample record to Samvera's direct and minted objects mapping.

I'm adding everyone as reviewers, but we really only need one.

This won't appear in the update docs until this is merged.

@mlhale7 @mathewjordan @CanOfBees


markpbaggett:
@mathewjordan I've added width properties to both figures.  This will ensure that the img tags are wrapped in anchors.

mathewjordan (approved):
@markpbaggett Looks good now.

mlhale7 (commented):
@markpbaggett - thanks for this. One thing I wasn't sure on was if you intentionally left the relatedItem/identifier string out (MS.1324). I realize the ARK will get us to this finding aid, but I thought we'd likely retain the string too (especially because not everything has an ARK). I'll save any other questions for our meeting tomorrow to discuss together.

markpbaggett:
@mlhale7, I just missed this. 

The identifier in the complex option should be an identifier on utkphysicalcollections:1. 

The direct mappings option is more up to debate.  The instructions say, "Use for an identifier corresponding to a parent item that the item being described belongs to. [Type] should be replaced with the corresponding identifier type abbreviation from Library of Congress ​Standard IdentifierSchemes​." But in there examples, they model identifiers for `relatedItem[@type="otherVersion"]` but don't really show any other examples.  Putting a literal that is associated with a parent object on every single child seems wrong to me unless you are able to find a predicate that can express the relationship.

PR #4, opened 2020-07-08
Add examples of MODS notes.

markpbaggett:
This pull request does a few things:

1. Fixes the warnings from islandora_tulip_tree.rst that were causing local builds to break.
2. Tracks documents in top_level_elements.
3. Adds examples of all our notes with a few notes about our notes.

We will review these examples and make decisions about each at our next meeting.

@CanOfBees @mlhale7 @mathewjordan

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -19,6 +19,17 @@ This section contains materials for our final report.
 
    contents/*
 
+Top Level Elements
+------------------
+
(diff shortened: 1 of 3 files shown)


mlhale7 (approved):
Thanks for these @markpbaggett - they look "good" to me. I think @type="first line" and @type="First line" is the most disappointing. I'm going to go ahead and approve this, but please everyone else take a look.

PR #9, opened 2020-07-23
Note

markpbaggett:
This pull request does several things:

1. It gets part 4 of contents in line with the [official sphinx heading sytleguide](https://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html).  While I don't think this is necessary for all documents, it's critical for anything that is this big.
2. Adds namespaces used to this point.
3. Updates the mapping doc with decisions about note.
4. Expands the Toctree out to 2 levels for Contents.

It does not address decisions about these XPATHS, as I don't think we made any:

* `mods:note[@displayLabel="use and reproduction"]`
* `mods:note[@displayLabel="Transcribed from Original Collection"]`
* `mods:note[@displayLabel="Project Part"]`
* `mods:note[@displayLabel="Local Rights"]`
* `mods:note[@displayLabel="Attribution"]`


mlhale7 (approved):
@markpbaggett - Thanks for this. In the table you added, I appreciate the range column. This seems like a good way to handle describing the types of values that are valid.

PR #11, opened 2020-07-23
Add classification examples.

markpbaggett:
This adds all the classification examples from our Islandora metadata.

### DIFF
--- a/docs/source/top_level_elements/classification.rst
+++ b/docs/source/top_level_elements/classification.rst
@@ -1,0 +1,25 @@
+note
+====
+
+About
+-----
+
+This section describes all the different types of classifications that we have in our Islandora repository right now.
+
+No authority
+------------
+
+A few records have classifications with no authorities like `tenngirl:977 <https://digital.lib.utk.edu/collections/islandora/object/tenngirl:977/datastream/MODS>`_
+
+.. code-block:: xml
+
+    <classification>LD5296 .W6</classification>
+
+With an authority
+-----------------
+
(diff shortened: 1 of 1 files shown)


mlhale7 (commented):
@markpbaggett For classification, I had a value related concern that I wanted to mention. Some values include dates and others do not. For instance LD5296 .W6 and LD5291.8 .T67 1966. Technically classification is different than a shelf locator or call number and the year should not be included on this value. I'll open a ticket for this.

Otherwise, I wanted to mention that technically all classification values follow lcc even if this is not identified with an authority. I can open a ticket to fix this. Let me know if this is something we could also just knowingly map over (all values should be associated with lcc) rather than taking the time to do the metadata edits.

Let me know if there's anything we should incorporate into this doc before merging and if you have any suggestions.

markpbaggett:
@mlhale7, I added a commit but am not sure that's good enough.  Let me know what you think.

mlhale7 (commented):
Maybe change the text to the following?

Some records have values with years that need to be removed because these values aren't consistent with classification values. The values present currently are shelfLocators rather than purely subject-focused values. Islandora's recommendation to map classification values to dc:subject highlights this focus. An example record that requires edits is `harp 1 <https://digital.lib.utk.edu/collections/islandora/object/harp%3A1/datastream/MODS>`_

If you could also note in the "No authority" section that the values follow lcc despite not having this attribute, that would be helpful too.

markpbaggett:
Okay, done.

mlhale7:
@markpbaggett - thanks for the edits.

PR #12, opened 2020-07-23
Add part evaluation.

markpbaggett:
This adds all the occurrences of parts from our current Islandora metadata.


mathewjordan (approved):
Builds successfully.

PR #13, opened 2020-07-28
Add titleInfo analysis

mathewjordan:
**Jira Issue**: [DIGITAL-583](https://jirautk.atlassian.net/browse/DIGITAL-583)

This PR includes an analysis of all of the titleInfo top level element.

## How should this be tested?

Please check for any formatting issues before merging. Also note if you can think of an additional use case that is not represented in the document.

## Notes

I did include sample turtle attempts. These were meant to inform discussion and are not expected to be finalized mapping.

### DIFF
--- a/docs/source/top_level_elements/titleInfo.rst
+++ b/docs/source/top_level_elements/titleInfo.rst
@@ -19,6 +19,12 @@ https://digital.lib.utk.edu/collections/islandora/object/roth:5342/datastream/MO
         <title>Norris Dam</title>
     </titleInfo>
 
+.. code-block:: turtle
+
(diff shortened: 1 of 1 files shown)


mlhale7:
@mathewjordan - the formatting works in pycharm for me. Noting for others, I am still hoping to remove @displayLabel="Cover Title" as well as the "Also Known As" titles for colloquy. If I do so, I will open up a new PR removing the examples for these so that we don't get confused about what we need to map. Ultimately it will be much less time intensive for us to handle some of these things without metadata remediation and simply map them to a more standard field.

mlhale7 (approved):

markpbaggett (approved):
Looks good to me :+1:

PR #14, opened 2020-07-29
Remove note[@type="provenance"] and add decisions from 7-30 meeting.

mlhale7:
**Jira Issue**: [DIGITAL-780](https://jirautk.atlassian.net/browse/DIGITAL-780)

Removes mods:note[@type="provenance"] from our example analysis since these values have been deleted from our existing metadata. Additionally, it adds decisions on note and internetMediaType values made during our 7-30 meeting.

## How should this be tested?

Please check formatting and evaluate the accuracy of the written decisions.

## Interested parties

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -83,6 +83,54 @@ originInfo
 physicalDescription
 ===================
 
+extent
+------
+
+Use Case
+^^^^^^^^
+
+Justification
+^^^^^^^^^^^^^
+
+Xpath
+^^^^^
+
+Decision
+^^^^^^^^
+
+
+
+
+internetMediaType
+-----------------
+
+Use Case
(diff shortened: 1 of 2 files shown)


markpbaggett:
@mlhale7, maybe we can talk about this tomorrow, but I kind of feel the analysis pages are there to reflect what was there for initial analysis.  I don't think we need to go back and modify the analysis pages.  In some ways, I think that anything from the analysis that is going to be around should be in the final mapping document, and we should remove the analysis pages prior to final release.

mlhale7:
@markpbaggett - sure, let's talk more tomorrow. I guess I'm thinking that this information isn't "going to be around" as it has already been removed. Therefore I felt it was confusing to include a decision and justification for an xpath that no longer exists in our final mapping.

markpbaggett (approved):
Looks good to me. :+1:

PR #15, opened 2020-07-30
titleInfo mapping

mathewjordan:
This pull request adds mapping decisions for titleInfo based on decisions made during the 7/28 Mods to RDF meeting between Meredith Hale and Mat Jordan. This one was pretty simple, but we have made a couple of key decisions.

- For cleanliness and consistency `displayLabel` won't be used to describe titles. Meredith intends to clean this up prior to migrating. If not we will simply exclude the defining of displayLabels during the migration process.
- We will **NOT** used `[` or `]` wrapping characters to define titles as supplied.
- When `<partName>` subelements are present in MODS, we will use commas to indicate enumeration of an object during the transcription to RDF.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -284,7 +284,7 @@ An object with a two titleInfo elements and one having an attribute of type="alt
 
 Justification
 ^^^^^^^^^^^^^
-tbd
+Keeping direct mapping simple.
 
 Xpath
 ^^^^^
@@ -293,6 +293,8 @@ Xpath
 
 Decision
 ^^^^^^^^
+
+titleInfo elements with @type="alternative" will defined as dcterms:alternative
 
 `Example record from pcard00:100233 <https://digital.lib.utk.edu/collections/islandora/object/pcard00:100233/datastream/MODS/>`_
 
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -68,6 +68,245 @@ identifier
 titleInfo
 =========
 
++-----------------------------------+----------------+-------------------+-------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


mlhale7 (changes requested):
@mathewjordan - thanks for sharing this. I think adjusting the wording of the decision for "Multiple titleInfo elements with one having a supplied attribute of yes" might be helpful. I'd change "In cases where supplied="yes" are present for one titleInfo element the titleInfo/title element will be used as dcterms:title." to "In cases where supplied="yes" are present for one titleInfo element the mods:titleInfo[@supplied]/mods:title value will be used as dcterms:title." The way the first one is phrased makes it sound like we are using title value that isn't supplied.

Also for the nonSort sub-element, can we change the decision to "The string contained within the nonSort element will be prepended to the title value" (line 259).

There's also a small typo within the alternative title use case section. Please remove "a" in "An object with a two" (line 283).

Finally, you duplicated the alternative title heading on line 318 though the decisions below refer to displayLabels. Can you update this?

mathewjordan:
@mlhale7 @markpbaggett I've addressed the changes requested by Meredith. I'd like to get this looped into master if no other issues are present.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -175,7 +175,7 @@ Xpath
 Decision
 ^^^^^^^^
 
-In cases where supplied="yes" are present for one titleInfo element the titleInfo/title element will be used as dcterms:title.
+In cases where supplied="yes" are present for one titleInfo element the mods:titleInfo[@supplied]/mods:title value will be used as dcterms:title.In cases where supplied="yes" are present for one titleInfo element the mods:titleInfo[@supplied]/mods:title value will be used as dcterms:title.
```
mlhale7: @mathewjordan - Am I misreading this, or is the same sentence copied twice?

mathewjordan:
@mlhale7 Sorry about that. I've removed the duplicate statement. Thanks for catching.

PR #16, opened 2020-08-07
Add decisions and revisions for physicalDescription from 8-4-2020 mee…

mlhale7:

This PR adds decisions made verbally in our 8-4 and 8-7 meetings about the top-level element physicalDescription.

## What's new?

Use cases, Xpaths, justifications, and decisions were added for the following:
mods:physicalDescription/mods:note
mods:physicalDescription/mods:extent
mods:physicalDescription/mods:extent[@unit="pages"]
mods:physicalDescription/mods:form[@type="material"]

## How should this be tested?

Are any of the justifications not articulated properly or is the rst not formatted correctly?

## Interested parties

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -213,10 +213,10 @@ Decision
 
 .. code-block:: turtle
 
-    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
+    @prefix rdau: <http://rdaregistry.info/Elements/u/> .
 
     <https://example.org/objects/1>
-        http://rdaregistry.info/Elements/u/P60550 "3 1/4 x 5 inches" .
+        rdau:P60550 "3 1/4 x 5 inches" .
 
 extent - @unit
 --------------
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -216,8 +216,7 @@ Decision
     @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
 
     <https://example.org/objects/1>
-        rdau:extent.en "3 1/4 x 5 inches" .
-rdau:extent
+        http://rdaregistry.info/Elements/u/P60550 "3 1/4 x 5 inches" .
 
 extent - @unit
 --------------
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -118,21 +118,143 @@ We have decided to not migrate these values as is justified above. Here's an `ex
 
     <digitalOrigin>born digital</digitalOrigin>
 
+note
+----
+
+Use Case
+^^^^^^^^
+Two collections, the Botanical Photography of Alan S. Heilman and the William Derris Film Collection, include <note> elements
+within <physicalDescription>. These values are of two types. The majority of the values communicate camera settings for the
+Heilman collection, while a smaller number of values share the "Film type" that was used to produce the print that was
+digitized. Below is a small sample of these values:
+
+1. Camera setting: 7@50 on 25; with filter
+2. 0.18x magnification, 100 Velvia
+3. Film type: Kodachrome Transparency
+4. zoomA -> 70 [A], Auto f16E100s
+5. Film type: GEMounts
+
+These values are somewhat problematic because they do not describe the digitized resource, but instead provide information about
+the process that created these resources. This is useful information to know, but isn't tied directly to the resource, making
+the inclusion of the values within physicalDescription inaccurate.
+
+Justification
+^^^^^^^^^^^^^
+Since we do not use mods:physicalDescription/mods:note regularly, it would streamline our data if these values could be
+appropriately placed elsewhere. I attempted to match film type values ("GEMounts" and "Kodachrome Transparency") with AAT
+terms, but wasn't able to find anything appropriate for "GEMounts." The accuracy of some of this information is questionable
+(for instance, GEMounts are likely a brand instead of a film type), but without access to the actual materials during the quarantine, it's
+impossible to make an informed judgement on what should be changed. To retain this contextual information that might
+prove useful to researchers interested in photographic processes and techniques, it seems best to simply put these values
+in a generic note field. If additional attention can be given to these two collections in the future, we can remediate
+the metadata following migration with the benefit of having access to the physical materials.
+
+Xpath
+^^^^^
+mods:physicalDescription/mods:note
+
+Decision
+^^^^^^^^
+All values will be moved to a generic note field.
+
+`Example record - derris:879 <https://digital.lib.utk.edu/collections/islandora/object/derris%3A879/datastream/MODS/view>`_
+
+.. code-block:: xml
+
+    <physicalDescription>
+        <form authority="aat" valueURI="http://vocab.getty.edu/aat/300127478">transparencies</form>
+        <digitalOrigin>digitized other analog</digitalOrigin>
+        <note>Film type: GEMounts</note>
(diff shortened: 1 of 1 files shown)


docs/source/contents/4_draft_mapping.rst
```diff
@@ -118,21 +118,143 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
+
+    <extent>3 1/4 x 5 inches</extent>
+
+.. code-block:: turtle
+
+    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
+
+    <https://example.org/objects/1>
+        rdau:extent.en "3 1/4 x 5 inches" .
+rdau:extent
```
markpbaggett: I think this property was copied and pasted here accidentally.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -118,21 +118,143 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
+.. code-block:: xml
+
+    <extent>3 1/4 x 5 inches</extent>
+
+.. code-block:: turtle
+
+    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
+
+    <https://example.org/objects/1>
+        rdau:extent.en "3 1/4 x 5 inches" .
```
markpbaggett: See the message in my review.  Also, the turtle lexer does not like namespaced URIs like this. If we don't change properties, I think we should just use the full URI.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -118,21 +118,143 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
+.. code-block:: xml
+
+    <extent unit="pages">4</extent>
+
+.. code-block:: turtle
+
+    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
+
+    <https://example.org/objects/1>
+        rdau:extent.en "4 pages" .
```
markpbaggett: See the message in my review. Also, the turtle lexer does not like namespaced URIs like this. If we don't change properties, I think we should just use the full URI.

markpbaggett (commented):
@mlhale7, thanks for putting this together. I think that rdau adds some complexities that we need to revisit and perhaps document in our philosophy.

When we discussed rdau, we decided that although it did not adhere to best practices for RDF and content negotiation in general that we would make an exception because it was important to separate extent into its own rdf property. To do this, we selected this rdf property:  `http://rdaregistry.info/Elements/u/#extent.en`. A few things about this property that I've noticed:

1. Even though rdau is not content negotiable, they include the RDF as a downloadable attachment. The actual rdf property listed here is <http://rdaregistry.info/Elements/u/P60550>.  We took the other URI from the Samvera working doc, which in my opinion is wrong. <http://rdaregistry.info/Elements/u/P60550> has a <http://metadataregistry.org/uri/profile/regap/lexicalAlias> with the value of <http://rdaregistry.info/Elements/u/extent.en>.
2. You'll notice that even the object of <http://metadataregistry.org/uri/profile/regap/lexicalAlias> doesn't match our examples.

I'm still okay with using rdau if others think we should, but I don't think we've chosen the right property and should reevaluate.  If we decide to use this property, I think we should justify it.

markpbaggett:
One more note:

rdau states:

```All RDA URIs have both an immutable canonical form and a "readable", lexical form, which is subject to change (changes will be redirected).```

This is not true, and our example is proof of that.

**Comment by mlhale7** [comment]
@markpbaggett - I believe I addressed your comments with the newest commit. Let me know if I've misunderstood.

File: docs/source/contents/4_draft_mapping.rst [hunk 4]
```diff
@@ -118,21 +118,142 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
+^^^^^^^^
+`Example record - knoxgardens:125 <https://digital.lib.utk.edu/collections/islandora/object/knoxgardens%3A125/datastream/MODS/view>`_
+
+.. code-block:: xml
+
+    <extent>3 1/4 x 5 inches</extent>
+
+.. code-block:: turtle
+
+    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
```
markpbaggett: Change the URI to : <http://rdaregistry.info/Elements/u/>

docs/source/contents/4_draft_mapping.rst
```diff
@@ -118,21 +118,142 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
+
+    <https://example.org/objects/1>
+        http://rdaregistry.info/Elements/u/P60550 "3 1/4 x 5 inches" .
```
markpbaggett: @mlhale7, the lexer didn't like the "." in the namespaced uri.  Can you change this to:

```
rdau:P60550
```
or

```
<http://rdaregistry.info/Elements/u/P60550>
```

docs/source/contents/4_draft_mapping.rst
```diff
@@ -118,21 +118,142 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
+.. code-block:: xml
+
+    <extent unit="pages">4</extent>
+
+.. code-block:: turtle
+
+    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
+
+    <https://example.org/objects/1>
+        rdau:extent.en "4 pages" .
```
markpbaggett: @mlhale7, the lexer didn't like the "." in the namespaced uri.  Can you change this to:

```
rdau:P60550
```
or

```
<http://rdaregistry.info/Elements/u/P60550>
```

docs/source/contents/4_draft_mapping.rst
```diff
@@ -118,21 +118,142 @@ We have decided to not migrate these values as is justified above. Here's an `ex
...
 ^^^^^^^^
+`Example record - colloquy:202 <https://digital.lib.utk.edu/collections/islandora/object/colloquy%3A202/datastream/MODS/view>`_
+
+.. code-block:: xml
+
+    <extent unit="pages">4</extent>
+
+.. code-block:: turtle
+
+    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .
```
markpbaggett: Change this to:

```
@prefix rdau: <http://rdaregistry.info/Elements/u/> .
```

markpbaggett (changes requested):

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks - I didn't realize I completely left off the carrots. I think the other option is cleaner though and will commit that.

mlhale7:
@markpbaggett - thanks for clearing up my misunderstandings. I've committed your suggestions.

markpbaggett (changes requested):
> @markpbaggett - thanks for clearing up my misunderstandings. I've committed your suggestions.

Meredith, we still have a problem:

1. 253 - 256 looks like this:

```
.. code-block:: turtle

    @prefix rdau: <http://rdaregistry.info/Elements/u/#> .

        rdau:extent.en "4 pages" .
```

but it should look like this:

```
.. code-block:: turtle

    @prefix rdau: <http://rdaregistry.info/Elements/u/> .

        rdau:P60550  "4 pages" .
```

docs/source/contents/4_draft_mapping.rst
markpbaggett: This is still an issue.  See below.

mlhale7:
@markpbaggett - apologies for completely overlooking that the same issue needed to be changed twice. Hopefully I've addressed everything now.

markpbaggett (approved):

PR #19, opened 2020-08-20
Add draft philosophy.

markpbaggett:
What Does This Do
===============

Converts our philosophy / best practices Google doc to a restructured text document.

If we need to move this around somewhere before publication we can use ```git mv```.

### DIFF
--- a/docs/source/contents/3_philosphy.rst
+++ b/docs/source/contents/3_philosphy.rst
@@ -42,8 +42,8 @@ specifically enable a user to agnostically request RDF from a URI and return RDF
 knowledge about a particular vocabulary or technical implementation. In other words, a user or application should not
 have to concatenate a string to the end of a URI in order to get back RDF.
 
-Principle Three: Adhere to rdfs:domain and rdfs:range Properties
-================================================================
+Principle Three: Adhere to "rdfs:domain" and "rdfs:range" Properties
+====================================================================
 
 When choosing RDF properties, the working group recognizes that the property in question may have rules for the subjects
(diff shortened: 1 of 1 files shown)


docs/source/contents/3_philosphy.rst
```diff
@@ -0,0 +1,97 @@
...
+When choosing RDF properties and classes, the working group attempts to select URIs that are not only dereferenceable
+but adhere to the principles of content negotiation. In order to be dereferenceable, the URI should return information
+about the URI when queried. This aligns with Tim Berners-Lee’s third principle by being able to “look up the properties
+and classes one finds in data, and get information from the RDF, RDFS, and OWL ontologies including the relationships
+between the terms in the ontology.” Additionally, the URIs should adhere to the principles of content negotiation and
+specifically enable a user to agnostically request RDF from a URI and return RDF without having to possess any special
+knowledge about a particular vocabulary or technical implementation. In other words, a user or application should not
+have to concatenate a string to the end of a URI in order to get back RDF.
+
+Principle Three: Adhere to rdfs:domain and rdfs:range Properties
```
mlhale7: @markpbaggett - it looks like having two colons here is making it think this is a field list (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#rst-field-lists)? Not sure if this is truly a problem.

markpbaggett: @mlhale7, let's wrap them then. I'll add a commit.

mlhale7:
@markpbaggett - this looks good to me.

PR #21, opened 2020-08-25
Correct missed problems from July and August.

markpbaggett:
What Does This Do
===============

Corrects a bunch of silent errors (some of which suddenly are no longer silent???) from the past few months.  Also, attempts to make some things a little more consistent stylistically.


mlhale7 (approved):
@markpbaggett - thanks for this. I had noticed that I wasn't always tabbing over with code blocks and attempted to address some of these. I appreciate you catching the rest. Seeing some of the errors is helpful for me going forward.

PR #22, opened 2020-08-25
Build on push

markpbaggett:
What Does This Do
===============

Adds an action to test Sphinx builds on push regardless on branch.  The build is based on our [make file](https://github.com/UTKcataloging/mods_to_rdf/blob/master/docs/Makefile) for building out html docs only.  If it builds without error, the build passes.  If it doesn't, fails. **Note**: our make file states to treat warnings as errors.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,16 @@
+name: Test Builds on Push
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+      - name: Install Sphinx
+        run: sudo apt-get install python3-sphinx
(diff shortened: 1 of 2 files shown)


mlhale7 (approved):
I reviewed the details of the checks that run. I agree that warnings can be treated as errors. @CanOfBees & @mathewjordan , if you have any comments or concerns feel free to share them.

markpbaggett:
@mlhale7, technically we've treated warnings as errors since the beginning in our application, but readthedocs doesn't use our makefile.  I couldn't figure out a way to change how they want to deploy and build  things in their infrastructure, so we've been building on warnings even though you couldn't build if you built locally.  This PR will make it so we know if our more restrictive build instructions fail.

mlhale7:
Merging assuming there are no additional comments - thanks Mark.

PR #24, opened 2020-08-28
Add language analysis.

mathewjordan:
This adds analysis and proposed turtle for language from our current Islandora metadata.

**Note:** the primary question I came across was what to with the items that had a language term code of **fra** (2 items in volvoices) and how these should be handled in relation to those with a common code of **fre**.

- volvoices:9928
- volvoices:11719

### DIFF
--- a/docs/source/top_level_elements/language.rst
+++ b/docs/source/top_level_elements/language.rst
@@ -18,7 +18,7 @@ https://digital.lib.utk.edu/collections/islandora/object/tatum%3A188/datastream/
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/eng> .
 
@@ -35,7 +35,7 @@ https://digital.lib.utk.edu/collections/islandora/object/ekcd:9/datastream/MODS/
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/eng> .
 
@@ -57,7 +57,7 @@ https://digital.lib.utk.edu/collections/islandora/object/utsmc:725/datastream/MO
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1>
         dcterms:language <http://id.loc.gov/vocabulary/iso639-2/fre> ;
@@ -78,7 +78,7 @@ https://digital.lib.utk.edu/collections/islandora/object/egypt:59/datastream/MOD
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1>
         dcterms:language <http://id.loc.gov/vocabulary/iso639-2/eng> ;
@@ -97,7 +97,7 @@ https://digital.lib.utk.edu/collections/islandora/object/tdh:911/datastream/MODS
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/zxx> ;
 
@@ -114,7 +114,7 @@ https://digital.lib.utk.edu/collections/islandora/object/tdh:911/datastream/MODS
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/zxx> ;
 
@@ -138,7 +138,7 @@ https://digital.lib.utk.edu/collections/islandora/object/volvoices:9928/datastre
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/fre> ;
 
@@ -162,7 +162,7 @@ to migrate.
 
 .. code-block:: turtle
 
-    @prefix dcterms: <http://purl.org/dc/terms/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1>
         dcterms:language <http://id.loc.gov/vocabulary/iso639-2/eng> ;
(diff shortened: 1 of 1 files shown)


mlhale7:
@mathewjordan - thanks for opening this.

I had a few brief comments. I'm not sure that we need to distinguish "zxx" / "No linguistic content" as a separate section. Since "zxx" comes from the same code list as "eng" (), we are treating them the same. I realize most codes are human readable / intuitive and "zxx" isn't, but it's the same. I think having this section unnecessarily complicates the final mapping document (especially if an external entity is using it). In some ways we could even consider simplifying the use cases to focus just on the authority instead of the text versus code - @authority of "iso639-2b". I'd be interested to hear what others think about this.

Second, I wanted to state that the collection-level record is likely something we are going to want to migrate since this is how we provide access to the collection in Alma/Primo. If we need to nest each languageTerm element within a language element to make it consistent with the structure of other records, this is something I can do.

Finally, this is not essential for this PR, but we should talk more about how we want to present links to objects within the mapping. I think it's important for the link to be live so that those looking at the document can click and easily link out. Generally I've followed the pattern of "`Example record - namespace:### <link>`_" Again, this is thinking about the final document and its usability.

I agree with all of the RDF decisions.

mlhale7:
I'm realizing this is just the analysis rst (not the final mapping), but you have basically everything we'd need for the final mapping. Sorry that some of my comments were on the final mapping doc, but it would be great to not create extra work if not needed. I think different people are approaching analysis differently and that we should talk more today about this.

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    <language>
+        <languageTerm type="code">fra</languageTerm>
+    </language>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/fre> ;
```
markpbaggett: You need a closing period instead of a semicolon.

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    <language>
+        <languageTerm type="code" authority="iso639-2b">zxx</languageTerm>
+    </language>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/zxx> ;
```
markpbaggett: You need a closing period instead of a semicolon.

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    <language>
+        <languageTerm authority="iso639-2b" type="text">No linguistic content</languageTerm>
+    </language>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:language <http://id.loc.gov/vocabulary/iso639-2/zxx> ;
```
markpbaggett: You need a closing period instead of a semicolon.

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
+language
+========
+
+About
+_____
+This section describes all the different types of language elements that we have in our Islandora repository right now.
+
+language has languageTerm subelement with an @authority of "iso639-2b" and a @type of "text"
+--------------------------------------------------------------------------------------------
```
markpbaggett: Headings here look weird.  Is the intention of headings like this to be a subheading of about or a sibling heading?

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
...
+
+language has language subelement without a stated @authority and a @code value of "fra"
+---------------------------------------------------------------------------------------
+
+Two questions here along with my proposed:
+
+1. How do we handle any that have no stated authority?
+2. Two volvoices objects use "fra". Is there a difference between fra and fre as codes?
+
+https://digital.lib.utk.edu/collections/islandora/object/volvoices:9928/datastream/MODS/vieww
```
markpbaggett: There is an extra w at the end on this URI.

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
...
+
+
+
+language has language subelement without a stated @authority and a @code value of "fra"
+---------------------------------------------------------------------------------------
+
+Two questions here along with my proposed:
+
+1. How do we handle any that have no stated authority?
+2. Two volvoices objects use "fra". Is there a difference between fra and fre as codes?
```
markpbaggett, line 132: We should talk about this in justification once we create that.

docs/source/top_level_elements/language.rst
```diff
@@ -0,0 +1,172 @@
...
+        <languageTerm authority="iso639-2b" type="text">Italian</languageTerm>
+        <languageTerm authority="iso639-2b" type="text">Spanish</languageTerm>
+    </language>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1>
+        dcterms:language <http://id.loc.gov/vocabulary/iso639-2/eng> ;
```
markpbaggett, line 171: Nothing wrong here, but I'd simplify since we only have 1 property.  Since style guide doesn't cover this, I'm okay to take as is.

```
 @prefix dcterms: <http://purl.org/dc/terms/> .

        dcterms:language <http://id.loc.gov/vocabulary/iso639-2/eng> , <http://id.loc.gov/vocabulary/iso639-2/fre>, <http://id.loc.gov/vocabulary/iso639-2/ger>, <http://id.loc.gov/vocabulary/iso639-2/ita>, <http://id.loc.gov/vocabulary/iso639-2/spa> . 

```

markpbaggett (changes requested):
Some comments, thoughts, and questions interlaced with a few requested changes.

docs/source/top_level_elements/language.rst
mathewjordan: Addressed syntax issue in 6d1fdfe

docs/source/top_level_elements/language.rst
mathewjordan: Addressed hierarchy issues in 6d1fdfe, making subsubsections for my distinct cases

docs/source/top_level_elements/language.rst
mathewjordan: Addressed syntax issue in 6d1fdfe

docs/source/top_level_elements/language.rst
mathewjordan: Addressed syntax issue in 6d1fdfe

docs/source/top_level_elements/language.rst
mathewjordan: Thanks, addressed in 6d1fdfe

docs/source/top_level_elements/language.rst
mathewjordan: I'll move to that next, thanks.

docs/source/top_level_elements/language.rst
mathewjordan: I see, that makes sense.

mathewjordan:
@markpbaggett I've addressed your requested changes, merged from master (to get additional testing). Things are now passing.

markpbaggett (approved):

mlhale7 (approved):
@mathewjordan - thanks for this. I'll second the approval. When you commit to the final mapping doc you can consider some of the suggestions I incorrectly provided here about streamlining the number of categories.

PR #25, opened 2020-09-01
Turtle tester 

markpbaggett:
What Does This Do
===============

This pull request does several things:

1. Adds an integration test to ensure turtle code blocks are valid and meet our style requirements.
2. Fixes bad turtle in code blocks in all directories except working_docs.

Integration Test
============

This adds an integration test to check that .rst files containing turtle code blocks in doc/source/contents or doc/source/top_level_elements are valid.  It does not test those in doc/source/working_docs because we already have enough problems on our hands, and I assume we'll drop theses files at some point. Additionally, an exception is raised if there is a SPARQL-style prefix in a code block.

Exceptions are gathered over each file and all exceptions are reported before a final exception is raised.

Note
====

Until we address issues in: 4_draft_mapping.rst, titleInfo.rst, origininfo_dates.rst, and origininfo_publishing.rst, builds will fail.

I also strongly suggest that we merge no more code until we deal with these issues.


mlhale7 (approved):
@markpbaggett - thank you for this! I reviewed the edits to the mapping, title, and originInfo docs so that I can be more vigilant about not introducing similar issues in the future. It will be great to have this code to catch them when they do occur though!

markpbaggett:
One question:  Should this test run when you open a pull request?  If we don't add this, any branches that don't include these changes won't have automated tests run until after merge.

CanOfBees:
Re your question, @markpbaggett, I think having this applied when opening a  PR would be great.

markpbaggett:
I thought about this more. I'm not sure this will work on pull requests unless the pull request has this code.  Instead, let's just require a pull from master prior to opening the p.r.

mlhale7:
@markpbaggett - thanks for adding additional reporting so that we get information on more than the first fail.

PR #28, opened 2020-09-15
Add language mapping decisions

mathewjordan:
Jira Issue: https://jirautk.atlassian.net/browse/DIGITAL-847

### What does this Pull Request do?
This adds mapping decisions for the top-level element language.

### How should this be tested?
Check to make sure all the tests pass and the syntax is good. Also read the use cases, justifications, and decisions to see if you agree with the content and decisions made.

### Anything else?
For reference: [language analysis](https://utk-mods-to-rdf.readthedocs.io/en/latest/top_level_elements/language.html)

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -1225,17 +1225,18 @@ language
 |                                   |                |                   | value from a controlled vocabulary, such as ISO 639-2.                  |
 +-----------------------------------+----------------+-------------------+-------------------------------------------------------------------------+
 
-language
-------------------
+item has one language
+---------------------
 
 Use Case
 ^^^^^^^^
-Single instance of languageTerm where item language is known. Many of our resources will have one instance of language with a single subelement
-of languageTerm. The *type* attribute for *languageTerm* may be either **text** or **code**.
+Single instance of languageTerm where item language is known. Many of our resources will have one instance of a
+language element with a single subelement of languageTerm. The *type* attribute for *languageTerm* may be either
+**text** or **code**.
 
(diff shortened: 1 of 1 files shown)


markpbaggett (approved):
Looks good to me.

mathewjordan:
Anyone else have a review for this one? @CanOfBees @mlhale7

mlhale7:
If you could, it would be helpful to add syntax to make the links to the digital objects live. This will make it easy for users to look at the rest of the record if they'd like. This is also something that can be fixed in editing, but it will save time if we all consistently do this along the way.

mlhale7:
As I commented in your initial PR for analysis, I also don't think "No linguistic content" should be a separate category. All of the codes ("zxx" and "eng") come from the same vocabulary.

mathewjordan:
> If you could, it would be helpful to add syntax to make the links to the digital objects live. This will make it easy for users to look at the rest of the record if they'd like. This is also something that can be fixed in editing, but it will save time if we all consistently do this along the way.

From what I can see in my local `make html` output and in GitHub's interpretation, the URLs for objects are output as hyperlinks. If this is a suggestion to clean it up so that we have the same formatting, I'm fine with going that way.

Should they all follow this pattern?

**[Example record from sanborn:1194](https://digital.lib.utk.edu/collections/islandora/object/sanborn:1194/datastream/MODS/)**

mlhale7:
@mathewjordan - thanks for making that change for zxx values. I think it's fine to leave your links as they are for now. I think all four of us have slightly different approaches to this.

mlhale7 (approved):
Approving and merging.

PR #29, opened 2020-09-18
Add location analysis

mathewjordan:
This adds analysis and proposed turtle for **location** from our current Islandora metadata. 

There are a few instances that stumped me a bit.[ Item's with XML like the below](https://github.com/UTKcataloging/mods_to_rdf/blob/locationAnalysis/docs/source/top_level_elements/location.rst#physicalocation-with-holdingsimple-and-holdingexternal) leave me wondering what should be retained. Either way, I understand this is just a the analysis step of this element, and so, turtle does not need to be perfect yet. I expect we will review this further and make modifications following the 9/18 Mods to RDF meeting.

https://digital.lib.utk.edu/collections/islandora/object/volvoices:2199/datastream/MODS/view

```
    <location>
        <physicalLocation>University of Memphis. Special Collections</physicalLocation>
        <holdingSimple>
            <copyInformation>
                <shelfLocator>Manuscript Number 5</shelfLocator>
            </copyInformation>
        </holdingSimple>
        <holdingExternal>
            <holding xsi:schemaLocation="info:ofi/fmt:xml:xsd:iso20775 http://www.loc.gov/standards/iso20775/N130_ISOholdings_v6_1.xsd">
                <physicalAddress>
                    <text>City: Memphis</text>
                    <text>County: Shelby County</text>
                    <text>State: Tennessee</text>
                </physicalAddress>
            </holding>
        </holdingExternal>
    </location>
```

### DIFF
--- a/docs/source/top_level_elements/location.rst
+++ b/docs/source/top_level_elements/location.rst
@@ -8,8 +8,10 @@ This section describes all the different types of location elements that we have
 Distinct Cases
 --------------
 
-physicalLocation with explicit uri
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+physicalLocation with @valueURI
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We have 14302 where a physicalLocation has a valueURI attribute. We will use the URI in replacement of a literal in those cases.
 
 https://digital.lib.utk.edu/collections/islandora/object/egypt:79/datastream/MODS/view
 
@@ -26,8 +28,39 @@ https://digital.lib.utk.edu/collections/islandora/object/egypt:79/datastream/MOD
     <https://example.org/objects/1>
         relators:rps <http://id.loc.gov/authorities/names/no2017033007> .
 
+physicalLocation without @valueURI
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If valueURI is not present, we will use the string literal.
+
+https://digital.lib.utk.edu/collections/islandora/object/egypt:79/datastream/MODS/view
+
+.. code-block:: xml
+
+    <location>
+        <physicalLocation>Blount County Public Library</physicalLocation>
+        <holdingExternal>
+            <holding xsi:schemaLocation="info:ofi/fmt:xml:xsd:iso20775 http://www.loc.gov/standards/iso20775/N130_ISOholdings_v6_1.xsd">
+                <physicalAddress>
+                    <text>City: Maryville</text>
+                    <text>County: Blount County</text>
+                    <text>State: Tennessee</text>
+                </physicalAddress>
+            </holding>
+        </holdingExternal>
+    </location>
+
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+
+    <https://example.org/objects/1>
+        relators:rps "Blount County Public Library" .
+
 physicalLocation as string (special collections)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If we do not have a valueURI for the various strings encompassing University of Tennessee Libraries Special Collections, we would map these to the appropriate URI.
 
 https://digital.lib.utk.edu/collections/islandora/object/roth:4245/datastream/MODS/view
 
@@ -46,6 +79,8 @@ https://digital.lib.utk.edu/collections/islandora/object/roth:4245/datastream/MO
 
 physicalLocation as string (libraries)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Like above, if we do not have a valueURI for the various strings encompassing University of Tennessee Libraries, we would map these to the appropriate URI.
 
 **The University of Tennessee Libraries, Knoxville** 574 instances
 
--- a/docs/source/top_level_elements/location.rst
+++ b/docs/source/top_level_elements/location.rst
@@ -39,6 +39,7 @@ https://digital.lib.utk.edu/collections/islandora/object/roth:4245/datastream/MO
 .. code-block:: turtle
 
     @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
 
     <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/no2014027633> .
 
@@ -93,13 +94,14 @@ https://digital.lib.utk.edu/collections/islandora/object/scopes:1258/datastream/
         <shelfLocator>Box 5, Folder 8</shelfLocator>
     </location>
 
+In most cases, especially those under our purview, we will likely opt to drop box and folder number references.
+
 .. code-block:: turtle
 
     @prefix relators: <http://id.loc.gov/vocabulary/relators> .
     @prefix opaque: <http://opaquenamespace.org/ns/> .
 
-    <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/no2014027633> ;
-        opaque:locationShelfLocator "Box 5, Folder 8" .
+    <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/no2014027633> .
 
 physicalLocation with holdingSimple and holdingExternal
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -131,8 +133,7 @@ https://digital.lib.utk.edu/collections/islandora/object/volvoices:2199/datastre
     @prefix relators: <http://id.loc.gov/vocabulary/relators> .
     @prefix opaque: <http://opaquenamespace.org/ns/> .
 
-    <https://example.org/objects/1> relators:rps "University of Memphis. Special Collections" ;
-        opaque:locationShelfLocator "Manuscript Number 5" .
+    <https://example.org/objects/1> relators:rps "University of Memphis. Special Collections" .
 
 physicalLocation with displayLabel="Address"
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


mlhale7 (approved):
I approve of this as an analysis document showing what we have in our metadata.

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,208 @@
...
+        <url access="preview">https://digital.lib.utk.edu/collections/islandora/object/volvoices%3A9999/datastream/TN/view</url>
+    </location>
+
+.. code-block:: turtle
+
+    @prefix edm: <http://www.europeana.eu/schemas/edm/> .
+
+    <https://example.org/objects/1> edm:isShownAt <https://digital.lib.utk.edu/placeholder/shownat/uri> ;
+        edm:preview <https://digital.lib.utk.edu/placeholder/preview/uri> ;
+        edm:object <https://digital.lib.utk.edu/placeholder/object/uri> .
```
markpbaggett: This all should be dropped.  It's contextual and relates to our current system.  We would not want to migrate any of this, and in my opinion none of this is descriptive metadata and outside the scope of this document.

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,208 @@
...
+
+*I am not sold on whether retaining the Archives Collection string is necessary. I don't think the city and state are necessary if a URI is used instead of string.*
+
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+
+    <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/no2001080757> ;
+        dbo:collection "Archives Collection" .
```
markpbaggett: Are you suggesting that <https://example.org/objects/1> is an "Archives Collection?"

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,208 @@
...
+
+*This is one where I'm not sure where to go with.*
+
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+
+    <https://example.org/objects/1> relators:rps "Pi Beta Phi Fraternity" ;
+        opaque:locationShelfLocator "Box 36, Folder 14" .
```
markpbaggett: URI resolves as:


"Hmmmm...

The vocabulary or term you were looking for doesn't exist."

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,208 @@
...
+        </holdingExternal>
+    </location>
+
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+
+    <https://example.org/objects/1> relators:rps "University of Memphis. Special Collections" ;
+        opaque:locationShelfLocator "Manuscript Number 5" .
```
markpbaggett: URI resolves as:

"Hmmmm...

The vocabulary or term you were looking for doesn't exist."

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,208 @@
...
+        <shelfLocator>Box 5, Folder 8</shelfLocator>
+    </location>
+
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+
+    <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/no2014027633> ;
+        opaque:locationShelfLocator "Box 5, Folder 8" .
```
markpbaggett: URI resolves as:

"Hmmmm...

The vocabulary or term you were looking for doesn't exist."

markpbaggett (changes requested):
I guess this is a mix of requested changes and comments. Let me know if you want to discuss or if you have questions.

docs/source/top_level_elements/location.rst
mlhale7: We had talked about this in our meeting. I'd agree that all of the actual values we have here should be dropped as they will mean nothing in the new system. Mat I think was wondering if more generally we might want to include a field for edm:preview and edm:object as they relate to the new system. I'm not entirely convinced this needs to be in the descriptive metadata (since we won't have the identifiers to create the links until after the objects are initially ingested). I don't think our OAI-PMH process will change radically and I don't see a need to reference the object URL when the user is already viewing the object, but I'm open to talking more about including a place to map these types of values if we think they would be relevant.

docs/source/top_level_elements/location.rst
mathewjordan: This is a good catch. Samvera's MODS to RDF Mapping Recommendations used this in their recommendations and it caught me. Apparently, I did not read the footnote stating it did not exist yet. 

>  This predicate does not yet exist; see Analysis Process section for more information on OpaqueNamespace.

I will instead suggest that we drop shelfLocator instances in most cases. When absolutely necessary, **boxNumber**, **boxName**, **folderNumber**, and **folderName** are available, see: http://opaquenamespace.org/predicates.

docs/source/top_level_elements/location.rst
mathewjordan: @markpbaggett I was suggesting that is in the collection _Archives Collection_ at http://id.loc.gov/authorities/names/no2001080757 repository. Should we drop this?

docs/source/top_level_elements/location.rst
mathewjordan: Here's what I've come around to here:

_The URIs referenced are relative to our current system and would not be migrated. Translating this to RDF would be self-referential and not descriptive metadata. This can be dropped._

docs/source/top_level_elements/location.rst
mathewjordan: See https://github.com/UTKcataloging/mods_to_rdf/pull/29#discussion_r496734929

docs/source/top_level_elements/location.rst
mathewjordan: See https://github.com/UTKcataloging/mods_to_rdf/pull/29#discussion_r496734929

mathewjordan:
Hey @markpbaggett @mlhale7 @CanOfBees I've made changes, see comments above for additional context. This is ready for another round of review.

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,203 @@
...
+https://digital.lib.utk.edu/collections/islandora/object/scopes:1258/datastream/MODS/view
+
+.. code-block:: xml
+
+    <location>
+        <physicalLocation valueURI="http://id.loc.gov/authorities/names/no2014027633">University of Tennessee, Knoxville. Special Collections</physicalLocation>
+        <shelfLocator>Box 5, Folder 8</shelfLocator>
+    </location>
+
+In most cases, especially those under our purview, we will likely opt to drop box and folder number references.
```
mlhale7: Can we specify the exact collections this does not apply to so that we can drop accordingly? I know you have included some arrowmont records below that seem to show when we are keeping shelfLocator, but it would be good to be very specific. I can help come up with a list if you'd like.

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,203 @@
...
+        <physicalLocation>University of Tennesse Knoxville. Libraries</physicalLocation>
+    </location>
+
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+
+    <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/n80003889> .
+
+physicalLocation and shelfLocator
```
mlhale7, line 122: I have some concerns about making a decision on both physicalLocation and shelfLocator simultaneously. I can think of some instances in which we might want to keep the physicalLocation value, but not the shelfLocator.

docs/source/top_level_elements/location.rst
```diff
@@ -0,0 +1,203 @@
...
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+
+    <https://example.org/objects/1> relators:rps <http://id.loc.gov/authorities/names/no2014027633> .
+
+physicalLocation with holdingSimple and holdingExternal
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+https://digital.lib.utk.edu/collections/islandora/object/volvoices:2199/datastream/MODS/view
```
mlhale7, line 181: Particularly with the Volunteer Voices Collection, we are going to need to think about how physicalLocation works with recordContentSource. I had forgotten that this collection only lists the contributing institution in physicalLocation and not in recordContentSource (which differs from collections like cdf and egypt). The issue is making sure that the institution (Un. of Memphis in this case) gets credit in DPLA for sharing this item instead of us. As I have laid out the mapping in recordContentSource, we would get credit for this item.

docs/source/top_level_elements/location.rst
mathewjordan: @mlhale7 I'm not dropping physicalLocation here. We retain that in all cases.

docs/source/top_level_elements/location.rst
mathewjordan: Perhaps this is counter to the point of analysis, but I'm not sure if I can really breakdown exactly where that threshold exists. This isn't really an XPath decision as much as one that would need to be curated by someone with care for the records, the collections, and the physical repository they belong to. Because of that, I'm thinking of this analysis doc as a stepping stone before the mapping decisions if the cases (to drop shelfLocator literals or translate to RDF using opaquenamespace predicates) do come about, not what records will go into which pile.

At the same time, maybe I have missed something in my analysis that would be helpful to mapping.

I'll review my [spreadsheet](https://docs.google.com/spreadsheets/d/1ghf0sP2zX2mC9aEgs7JmvhY9L5fq5YQcNWQf7WJV0Ts/edit?usp=sharing) more thoroughly to see if there are cases I might be missing.

docs/source/top_level_elements/location.rst
markpbaggett: Sorry, misunderstood.

CanOfBees (approved):
@mathewjordan I think this looks good with these latest changes (shelf locator update/changes)

mathewjordan:
@markpbaggett @mlhale7 This feels like 10 years ago. Should have pinged you back last Tuesday. The requested changes have been made. Theses should reflect decisions discussed verbally during our _Name Day_ conversations.

mlhale7:
@CanOfBees @markpbaggett @mathewjordan - I have given my approval for the analysis doc. For the mapping we will need to include information on Volunteer Voices so that the location / contributor is mapped to edm:dataProvider. Let me know if I should expand upon my previous comment so this can be addressed here or if we should just address it in the mapping (my recommendation).

markpbaggett (approved):
:+1:

PR #30, opened 2020-09-22
Add relatedItem top level document 

CanOfBees:
Add the relatedItem analysis to our top level document set.


docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "Digital Collection: The Growth of Democracy in Tennessee: A Grassroots Approach to Volunteer Voices" .
+
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/1> .
```
markpbaggett: I feel somewhat strongly that we shouldn't migrate or make recommendations about things that are more structural than descriptive.  I feel it's outside the scope of this doc, but I'm open to hearing opposing view points. The reason I feel this way is that the system we choose must implement PCDM (it's a requirement for our next gen platform), and the system will choose a PCDM accepted way for how it models the relationship between the collection and the object. 

Again, this is out of scope of this doc in my opinion, but I also feel like if you're describing something as a PCDM collection and its relationship to an object (like on 28-29 and on 31) that one of these things should be there in order to support longterm interoperability:  

1. the object should be a member of the collection:

``` turtle

@prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
@prefix pcdm: <http://pcdm.org/models#> .

<https://example.org/collections/1> a pcdm:Collection ;
    rdfs:label "Digital Collection: The Growth of Democracy in Tennessee: A Grassroots Approach to Volunteer Voices" .

<https://example.org/objects/1> pcdm:memberOf <https://example.org/collections/1> .

```

2. the collection should state what objects it has:

``` turtle

<https://example.org/collections/1> a pcdm:Collection ;
    rdfs:label "Digital Collection: The Growth of Democracy in Tennessee: A Grassroots Approach to Volunteer Voices" ;
    pcdm:hasMember <https://example.org/objects/1> .

```

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label """Albert "Dutch" Roth Photograph Collection""" .
+
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/1> .
```
markpbaggett: Again, I feel somewhat strongly that we shouldn't migrate or make recommendations about things that are more structural than descriptive.  I feel it's outside the scope of this doc, but I'm open to hearing opposing view points. The reason I feel this way is that the system we choose must implement PCDM (it's a requirement for our next gen platform), and the system will choose a PCDM accepted way for how it models the relationship between the collection and the object. 

Again, this is out of scope of this doc in my opinion, but I also feel like if you're describing something as a PCDM collection and its relationship to an object (like on 52-53 and on 55) that one of these things should be there in order to support longterm interoperability:  

1. the object should be a member of the collection:

``` turtle

@prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
@prefix pcdm: <http://pcdm.org/models#> .

<https://example.org/collections/1> a pcdm:Collection ;
    rdfs:label """Albert "Dutch" Roth Photograph Collection""" .

<https://example.org/objects/1> pcdm:memberOf <https://example.org/collections/1> .

```

2. the collection should state what objects it has:

``` turtle

<https://example.org/collections/1> a pcdm:Collection ;
    rdfs:label """Albert "Dutch" Roth Photograph Collection""" ;
    pcdm:hasMember <https://example.org/objects/1> .

```

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+
+Note: do we have a feeling about syntax for literals with quotes? `'''Albert "Dutch" Roth'''` vs `"""Albert "Dutch" Roth"""`? Does it matter?
```
markpbaggett: IMHO, this does not matter.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "Botanical Photography of Alan S. Heilman" .
+
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/1> ;
+        dbo:collection "Botany Department Photographs, AR.0488" .
```
markpbaggett: I personally do not the that a physical collection should be minted.  What are the use cases here?  If we do mint, I do not think that making that object an instance of a pcdm:Collection is appropriate.  There is nothing explicit in RDF that expresses this, but it gets back to the intent of PCDM. From the website, PCDM _"is a flexible, extensible domain model that is intended to underlie a wide array of repository and DAMS applications. The primary objective of this model is to establish a framework that developers of tools (e.g., Samvera-based engines, such as Hyrax, Hyku, Sufia, and Avalon; Islandora; custom Fedora sites) can use for working with models in a general way, allowing adopters to easily use custom models with any tool. Given this interoperability goal, the initial work has been focused on structural metadata and access control, since these are the key actionable metadata."_  

If we don't need to mint physical / archival collections, the better property is `http://dbpedia.org/ontology/collection`.

If we need to mint, let's talk why and look for another property (not pcdm:collection).

markpbaggett: Ugh, misread.  Same as above.  Let's not worry about describing "repository collections" in this doc as that implementation may vary from repository to repository depending on how pcdm was implemented.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "Thompson Brothers Commercial Photographers" .
+
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/1> .
```
markpbaggett: Let's let relationships between the object and one of its collections be handled by the platform (don't migrate).

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "Children's Defense Fund" .
+
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/1> .
```
markpbaggett: Let's let relationships between the object and one of its collections be handled by the platform (don't migrate).

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+    @prefix rdau: <http://www.rdaregistry.info/Elements/u/#> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "From Pi Beta Phi to Arrowmont" .
```
markpbaggett: Let's let relationships between the object and one of its collections be handled by the platform (don't migrate).

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+
+    <https://example.org/collections/2> a pcdm:Collection ;
+        dcterms:isPartOf <https://example.org/collections/1> ;
+        rdfs:label "The Arrow of Pi Beta Phi" .
```
markpbaggett: What's the thought here? Would we model "the Arrow" as a separate collection or sub collection in the next platform?  If it is already a subcollection, I assume that it would be migrated and modelled structurally based on the repository platform / migration method. I personally see this as structural metadata that is outside the scope of this document if this subcollection is already a collection object in Islandora.

If it's not, let's discuss.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/2> ;
+        rdau:P60101 "The Arrow, Volume 27, Number 1" .
```
markpbaggett: Seems to me that volume and issue is the pertinent thing here.  @mlhale7, do we have this data elsewhere in our metadata record?

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix rdau: <http://www.rdaregistry.info/Elements/u/#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "Volunteer Voices" .
+
```
markpbaggett: Again, let's not even model this in this doc as it's structural.  If we'd migrate the collection objects metadata, would we even use rdfs:label?  I'd rather not even touch this and assume that the `titleInfo/title` of the collection object would be migrated just like any other object.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+    <https://example.org/objects/1> dcterms:isPartOf <https://example.org/collections/1> ;
+        rdau:P60101 "Harper's Weekly" .
+
```
markpbaggett: Is there any way to avoid rdau here? It's dereferenceable, but not content negotiable. 

The property is labeled as "is facsimile part of" with a definition of:
```
Relates  a resource to a larger resource of which a part is a discrete component  that exactly reproduces a related resource embodying a resource.
```

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:collection ;
+        rdfs:label """Albert "Dutch" Roth Photograph Collection""" .
```
markpbaggett: Same as above and throughout.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collections/1> a pcdm:Collection ;
+        rdfs:label "David Van Vactor Music Collection" .
```
markpbaggett: Same as above and throughout.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,702 @@
...
+
+.. code-block:: turtle
+
+    @prefix rdfs: <https://www.w3.org/TR/rdf-schema/> .
+    @prefix pcdm: <http://pcdm.org/models#> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/collection/1> a pcdm:Collection ;
+      rdfs:label "Sanborn Fire Insurance Maps Collection" ;
+      dcterms:abstract "The Sanborn Fire Insurance Maps are a large-scale representation of the growth and layout of American cities. The University of Tennessee Libraries' collection currently provides digital versions of the four earliest sets of Knoxville maps - 1884, 1890, 1903, and 1917." .
```
markpbaggett: Again, we'd migrate metadata attached to a collection object following this same doc.  I'd drop this unless it conflicts.

markpbaggett (changes requested):
A lot of thoughts and questions here. Let me know if anything is not clear.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett I follow your argument. I was thinking about capturing the descriptive relationship(s), but I'm seeing that this line of thought had some flaws. Thanks for this feedback.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett Again, I follow your argument. I was thinking about capturing the descriptive relationship(s), but I'm seeing that this line of thought had some flaws. Thanks for this feedback.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett Again, I follow and appreciate your argument. I was thinking about capturing the descriptive relationship(s), but I'm seeing that this line of thought had some flaws. Thanks for this feedback - it's very helpful.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett the volume/number information is be available in the abstract (it looks that data appears 316 times specifically in MODS for the Arrow).

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett I follow your argument. I was thinking about capturing the descriptive relationship(s), but I'm seeing that this line of thought had some flaws. Thanks for this feedback.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett we may be able to use `dbo:newspaper` but I'm not sure that fits the use case. The values for `relatedItem[@type='host'][@displayLabel='Is Part Of']/titleInfo/title` all point to print resources:

```
Memphis Press-Scimitar
Shelbyville Times-Gazette
Knoxville Journal
Nashville Tennessean
Chattanooga Times
Seattle Times
Georgian
Harper's Weekly
Oregonian
Frank Leslie's Illustrated Newspaper
New York World
Los Angeles Times
Evening Star
Rochester Herald
Chicago Tribune
Mt. Pleasant Record
San Francisco Chronicle
Los Angeles Examiner
Cincinnati Post
```

Edit: I think that almost all of these are papers, so maybe that would work. @markpbaggett I'd appreciate your understanding of the range/domain for `dbo:newspaper` though; see http://dbpedia.org/ontology/newspaper.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett thank you!

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett okay.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett I hear what you're saying.

CanOfBees:
@markpbaggett Thanks again for the detailed feedback.

docs/source/top_level_elements/relatedItem.rst
mlhale7: @markpbaggett - this is helpful for me to read as well. Would we still be able to share the collection string outside of the structural data?

Below are two ways in which we currently use the collection name string.

I don't think this is essential to keep, but note that we do currently used the MODS to display the collection name in the click for details section. I think the thought initially was to make it easy to navigate from a single item to the collection broadly (though you can just hit Browse at the top of the page too . . .). Personally, I don't think this is effective.

We do also share this as a string in DPLA, though currently nothing is done with it. I know Gretchen had worked on a white paper on including collection names in DPLA that doesn't seem to have come to fruition on the actual platform - https://docs.google.com/document/d/16r_px4GajLIOZMlyXyN0pJFfWpstJIKxHF3brgM4YOo/edit

docs/source/top_level_elements/relatedItem.rst
mlhale7: As we talked about briefly in our meeting today, I feel that while this information is useful it isn't consistent with the rest of our metadata (other newspaper items don't have a relatedItem like this) and the record includes this information in less structured ways in other elements throughout the record (abstract / publisher).

docs/source/top_level_elements/relatedItem.rst
mlhale7: I have always dropped mods:relatedItem/mods:abstract when remediating collections, so I'd say that dropping will also help us with consistency.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @markpbaggett I'm updating this to use `dcterms:bibliographicCitation` instead of the RDAU predicate.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,598 @@
...
+
+.. code-block:: xml
+
+    <relatedItem displayLabel="Digital Collection" type="host">
+      <titleInfo>
+        <title>Children's Defense Fund</title>
+      </titleInfo>
+    </relatedItem>
+
+
```
mlhale7: I'm assuming we'll also drop this as it is "synonymous with @displayLabel='Project'", but maybe adding one sentence would make this clearer (just saying this is the action that will be taken rather than referring to another use case).

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,598 @@
...
+        <title>Prints Collection</title>
+      </titleInfo>
+    </relatedItem>
+    <relatedItem displayLabel="Is Part Of" type="host">
+      <titleInfo>
+        <title>Harper's Weekly</title>
+      </titleInfo>
+    </relatedItem>
+
+:code:`[@displayLabel='Is Part Of']` may be sufficiently captured in other parts of the MODS (e.g. :code:`abstract`). Flip-flopping backwards, we should ignore this XPath.
```
mlhale7, line 157: Agree 100%. For clarity I wondered if it would be helpful to include the abstract in the XML code block above.

docs/source/top_level_elements/relatedItem.rst
```diff
@@ -0,0 +1,598 @@
...
+        <dateIssued>1948</dateIssued>
+      </originInfo>
+    </relatedItem>
+
+Decision: we'll ignore these and find an alternate way to express/represent featured items for a collection.
+
+relatedItem/location[physicalLocation]
+--------------------------------------
+
+This XPath appears once, in the record for the Charles Dabny collection; i.e. `collections:dabney <https://digital.lib.utk.edu/collections/islandora/object/collections:dabney/datastream/MODS/view>`_.
```
mlhale7: Tiny typo - "Dabny" versus "Dabney"

mlhale7 (approved):
@CanOfBees - I had some minor suggestions that might provide clarity, but I agree with all of the decisions you've outlined.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @mlhale7 thanks for this suggestion - I've added a note that hopefully helps clarify what we're doing here.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: @mlhale7 done! Good suggestion - thank you.

docs/source/top_level_elements/relatedItem.rst
CanOfBees: :facepalm:

mlhale7:
@CanOfBees - thanks for those corrections!

markpbaggett (approved):
:+1:

PR #31, opened 2020-09-22
Table of contents

mlhale7:
**Jira Issue**: [DIGITAL-592](https://jirautk.atlassian.net/browse/DIGITAL-592)

This PR adds examples of <tableOfContents> within UTK records and shares a mapping solution to communicate this information in RDF going forward. Note that I completed this PR to complement Bridger's work on relatedItem[@type="constituent"]. Punctuation within tableOfContent strings is not consistent across UTK collections, but I do not think this has any affect on accessibility to users. Due to the simplicity of this top-level element (only one Xpath) I did not feel that doing an analysis doc prior to adding decisions to the mapping doc was necessary.

## How should this be tested?

Do all tests pass? Do you agree with the mapping decision and justification?

## Additional Notes:

If we come to a final decision on how best to represent our Xpaths going forward, I can update this within this PR.

## Interested parties

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -608,6 +608,56 @@ abstract
 tableOfContents
 ===============
 
+Use Case
+^^^^^^^^
+
+The following collections include tableOfContents - David Van Vactor Music Collection, Tennessee Farm and Home Science,
+The Arrow of Pi Beta PhiThere are a total of 455 unique values. This element contains the names of individually titled
+parts that make up the larger resource. It is used to provide more detailed information on the content of a resource in
+a non-structured way. Note that punctuation separating part titles varies depending on the string values being separated.
+The following punctuation is present in UTK's tableOfContents elements: " -- ", " - ", and ";".
+
+Justification
+^^^^^^^^^^^^^
+
+This information aides keyword discovery by adding more text to the record and providing users with a listing of parts
+within the larger resource.
+
+Xpath
+^^^^^
+
+`mods:tableOfContents`
+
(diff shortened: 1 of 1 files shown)


markpbaggett (commented):
@mlhale7, thanks for this.  I have two questions that I'd probably know the answer to if I hadn't missed so many meetings lately:

1. Is there a reason why we just have the draft mapping and no top level element doc?  I assume this is due to you all deciding this was unnecessary.
2. What's the purpose of the table of contents data anyway?  I know we describe the contents, but is there a reason why this isn't done in a way to support interoperability (actually making it so that we could parse the data and make the book viewer show a navigable table of contents)?

mlhale7:
@markpbaggett - for question 1, I felt that this field was very similar to abstract in that it has no attributes or child elements. I think it is more effort than its worth to open two PRs for this element. For 2, I also feel that tableOfContents serves a very similar function to abstract as it is currently realized. I think it's helpful for this information to be included in a brief results display (I realize it currently isn't in Islandora - just note and abstract) so that users can quickly get a sense of what is present in the resource. It also simply adds more text to the descriptive metadata that might help with retrieval when more specific searches are used. It would be great for there to be more structure present in this top-level element that would support users being able to go to a specific part of a book, but this would require additional work on our end. I'm focusing on making sure what we have makes it in its entirety to the new platform - here that's just text.

markpbaggett (approved):

markpbaggett:
I approve.

CanOfBees (approved):
@mlhale7 I think this looks good; thanks for covering the different delimiters.

PR #33, opened 2020-10-08
Change the specifications for our test environment.

markpbaggett:
# What does this do

1. Changes the Pipenv instructions for installation to be less specific so that multiple versions of python can be used.  This will fix current builds and fails.
2. Adds a pull request template

# How Should This Be Tested?

Does the action here pass? Previously, things have been failing at the "Install Our Application with Pipenv" stage because of a mismatch between 3.8 and 3.9.

Look over the pull request template to ensure it encapsulates everything that is needed here.

# Additional Notes

Our current tests rely on https://github.com/VaultVulp/action-pipenv.  This action uses the python community's current Docker image associated with the tag `3`. Two days ago, the Python community changed the `3` designation to associate with Python 3.9 instead of Python 3.8.  Since our Pipfile states to build a Python 3.8 virtual environment but the container it's built on doesn't have Python 3.8, builds are failing.  I've opened a ticket with VaultVulp to discuss the possibilities of overriding the version of the Docker image since he already has that as a variable argument in his Docker build. In the meantime, for this repository, I think we can get around this by simply specifying Python  `3` rather than `3.8`.


mathewjordan (approved):
@markpbaggett This good from my eyes. The action `Install our application with pipenv` now passes on: [push, pull_request].
image

mathewjordan:
@markpbaggett I also think the template is good.

markpbaggett:
@mathewjordan, since we discussed on Slack, I'm going to go ahead and make a change here to auto assign reviewers on pull request.

mathewjordan (approved):
Coming back into this @markpbaggett after the re-request. This is working as expected. To other reviewing, this final commit was somewhat of a tack-on. You likely didn't receive the automated review request as you were already actively requested manually.

Thanks for adding. Just reviewing this gives me more insight into how easy these things are to implement. I'm now thinking about how I can combo github actions/buddy on other projects.

image

CanOfBees (approved):
@markpbaggett approved. Thanks for getting these fixes together.

mlhale7 (approved):

PR #34, opened 2020-10-15
Add typeOfResource analysis doc.

mlhale7:
**JIRA Issue**: [DIGITAL-582](https://jirautk.atlassian.net/browse/DIGITAL-582)

What Does this Do?
==================

This PR adds an analysis document for the top-level element typeOfResource.

How Should This Be Tested?
==========================

Make sure this passes syntax checks. Also provide input on suggested mappings as well as information that may not be accounted for but needs consideration.

Additional Notes?
==========================

Suggestions on the best way to handle multiple typeOfResource[collection="yes"] values are welcome. Potentially we could decide to map the values as "mixed material" instead of specifying all the different types of resources present in a collection-level record.

### DIFF
--- a/docs/source/top_level_elements/typeOfResource.rst
+++ b/docs/source/top_level_elements/typeOfResource.rst
@@ -1,0 +1,99 @@
+typeOfResource
+==============
+
+About
+-----
+
+This section describes the values and usage of <typeOfResource> within UTK metadata. Note that the manuscript attribute
+is not used in any record.
+
+typeOfResource with no attributes
+---------------------------------
+
+Most records currently have a typeOfResource value with no attributes. Depending on the item being described, it is possible
+for there to be multiple typeOfResource values in a single record. The Islandora Metadata Interest Group has carefully
+created a mapping to translate MODS typeOfResource values to dcterms resource types. A selection of the mapping is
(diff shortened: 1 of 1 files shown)


docs/source/top_level_elements/typeOfResource.rst
```diff
@@ -0,0 +1,99 @@
...
+.. code-block:: xml
+
+    <typeOfResource collection="yes">text</typeOfResource>
+    <typeOfResource collection="yes">still image</typeOfResource>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms: <http://id.loc.gov/vocabulary/resourceTypes/col> ;
```
markpbaggett: You're missing the type property on this line which is causing builds to fail.

markpbaggett (changes requested):
I think this looks good other than the one missing element.

docs/source/top_level_elements/typeOfResource.rst
mlhale7: Sorry! I thought I had addressed them all. I've committed this addition.

CanOfBees (approved):
@mlhale7 thanks for talking through this - approved.

markpbaggett (approved):
:+1:

PR #35, opened 2020-10-19
Add typeOfResource to mapping decisions.

mlhale7:
**JIRA Issue**: [DIGITAL-582](https://jirautk.atlassian.net/browse/DIGITAL-582)

What Does this Do?
==================

This builds upon the initial typeOfResource analysis PR and adds final decisions to the group mapping doc.

How Should This Be Tested?
==========================

Check to make sure that all tests pass. Check the content for consistency and justifiable property selections.

Additional Notes
================

One significant change from the analysis doc is a list of all collections that have missing typeOfResource values.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -1531,8 +1531,7 @@ Xpath
 Decision
 ^^^^^^^^
 
-During migration we will plan to add typeOfResource on a collection basis if possible. See the chart below for decisions.
-If not, these collections will be remediated either prior to or after migration.
+During or post migration we will plan to add typeOfResource on a collection basis if possible. See the chart below for decisions.
 
 +----------------------------+---------------------------------------------------+
 | collection PID             | dcterms:type                                      |
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -479,7 +479,6 @@ character will be used as glue when concatenating the strings.
     <https://example.org/objects/1> dcterms:title "Knoxville -- 1917, Sheet 56" .
 
 
-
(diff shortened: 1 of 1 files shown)


docs/source/contents/4_draft_mapping.rst
```diff
@@ -1399,6 +1398,164 @@ https://digital.lib.utk.edu/collections/islandora/object/utsmc:725/datastream/MO
...
+
+Xpath
+^^^^^
+
+:code:`not(mods:typeOfResource)`
+
+Decision
+^^^^^^^^
+
+During migration we will plan to add typeOfResource on a collection basis if possible. See the chart below for decisions.
```
markpbaggett: I think  we should state, "During or post migration ..."  I don't think it's guaranteed or even likely that we will be able to do this. 

I assume you included this as "a reminder" that we will need to do this at some point: before, during, or after migration.  I have no issues with including this if that's the case, but I don't think we should suggest we will do in during.

markpbaggett (changes requested):
I think we should change this one thing.  We can discuss if you all disagree.

docs/source/contents/4_draft_mapping.rst
mlhale7: I'd agree. I think I was trying to be less wishy-washy, but the fact of the matter is we don't know exactly when (or if) this work will be done and it will likely be post migration. I'll commit a change.

mlhale7: The sentence following this does clarify that post migration is possible, but yes, the way it is written it makes it seem that during migration is likely.

markpbaggett (approved):
Looks good!  :+1:

CanOfBees (approved):
@mlhale7 This looks good.

PR #36, opened 2020-10-22
Genre

CanOfBees:
What Does this Do?
==================

High level analysis and sample mappings of mods:genre.

How Should This Be Tested?
==========================

Do tests pass?

Additional Notes
================

@mathewjordan @markpbaggett I've already discussed this with @mlhale7, but would like your thoughts and feedback on the suggested mapping for `genre[\@authority="lcmpt"]`: the use of a different ontology (GND: https://d-nb.info/standards/elementset/gnd#) to express instrumentation. Thanks for your feedback.

### DIFF
--- a/docs/source/top_level_elements/genre.rst
+++ b/docs/source/top_level_elements/genre.rst
@@ -33,6 +33,12 @@ genre[@authority = 'aat']
 
     <genre authority="aat" valueURI="http://vocab.getty.edu/aat/300021140">Renaissance</genre>
 
+.. code-block:: turtle
+
+    @prefix edm: <http://www.europeana.eu/schemas/edm/> .
+
+    <https://example.org/object/1> edm:hasType <http://vocab.getty.edu/aat/300021140> .
+
 genre[@authority = 'lcsh']
 --------------------------
 
--- a/docs/source/top_level_elements/genre.rst
+++ b/docs/source/top_level_elements/genre.rst
@@ -18,7 +18,7 @@ Many of the records in Volunteer Voices have multiple `genre`s; e.g. `volvoices:
 
 .. code-block:: turtle
 
-    @prefix edm: <http://www.europeana.eu/schemas/edm/>
+    @prefix edm: <http://www.europeana.eu/schemas/edm/> .
 
     <https://example.org/object/1> edm:hasType ____ ; # get dct URI
         edm:hasType "letter" .
@@ -52,9 +52,9 @@ Used in the Archivision and Charlie Daniel collections; e.g. `cDanielCartoon:455
 
 .. code-block:: turtle
 
-    @prefix edm: <http://www.europeana.eu/schemas/edm/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
-    <https://example.org/object/1> edm:hasType <http://id.loc.gov/authorities/subjects/sh85040974> .
+    <https://example.org/object/1> dcterms:type <http://id.loc.gov/authorities/subjects/sh85040974> .
 
 and `archivision:1754 <>`:_
 
@@ -64,9 +64,9 @@ and `archivision:1754 <>`:_
 
 .. code-block:: turtle
 
-    @prefix edm: <http://www.europeana.eu/schemas/edm/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
-    <https://example.org/object/1> edm:hasType <http://id.loc.gov/authorities/subjects/sh85040974> .
+    <https://example.org/object/1> dcterms:type <http://id.loc.gov/authorities/subjects/sh85040974> .
 
 
 genre[@authority = 'lcgft']
@@ -80,9 +80,9 @@ Appears once in ekcd (ekcd:611), otherwise is present in Arrowmont, Van Vactor, 
 
 .. code-block:: turtle
 
-    @prefix edm: <http://www.europeana.eu/schemas/edm/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
-    <https://example.org/object/1> edm:hasType <http://id.loc.gov/authorities/genreForms/gf2014026131> .
+    <https://example.org/object/1> dcterms:type <http://id.loc.gov/authorities/genreForms/gf2014026131> .
 
 In 124 of the records in the VP Moore collection, with `@authority='lcgft'`, the `@authorityURI = 'http://id.loc.gov/authorities/genreForms'` is used; e.g. `vpmoore:50 <https://digital.lib.utk.edu/collections/islandora/object/vpmoore:50/datastream/MODS/view>`:_
 
@@ -92,9 +92,9 @@ In 124 of the records in the VP Moore collection, with `@authority='lcgft'`, the
 
 .. code-block:: turtle
 
-    @prefix edm: <http://www.europeana.eu/schemas/edm/>
+    @prefix dcterms: <http://purl.org/dc/terms/> .
 
-    <https://example.org/object/1> edm:hasType <http://id.loc.gov/authorities/genreForms/gf2014026173> .
+    <https://example.org/object/1> dcterms:type <http://id.loc.gov/authorities/genreForms/gf2014026173> .
 
 genre[@authority = 'lcmpt']
 ---------------------------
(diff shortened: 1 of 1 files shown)


docs/source/top_level_elements/genre.rst
```diff
@@ -0,0 +1,164 @@
...
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+    @prefix gnd: <https://d-nb.info/standards/elementset/gnd#> .
+
+    <https://example.org/object/1>
+        gnd:playedInstrument <http://id.loc.gov/authorities/performanceMediums/mp2013015074> ;
+        gnd:playedInstrument <http://id.loc.gov/authorities/performanceMediums/mp2013015342> ;
+        gnd:playedInstrument <http://id.loc.gov/authorities/performanceMediums/mp2013015748> ;
+        gnd:playedInstrument <http://id.loc.gov/authorities/performanceMediums/mp2013015540> ;
```
markpbaggett: This won't work in my opinion.  There are a few reasons why, but let's start by looking at our property.

https://d-nb.info/standards/elementset/gnd#playedInstrument has a `rdfs:range` of https://d-nb.info/standards/elementset/gnd#SubjectHeading and a `rdfs:domain` of https://d-nb.info/standards/elementset/gnd#DifferentiatedPerson.  

1. Technically, we have an issue with range here.  The object of our triple is a http://www.loc.gov/mads/rdf/v1#Authority, http://www.loc.gov/mads/rdf/v1#Medium, and http://www.w3.org/2004/02/skos/core#Concept. These inherit types like http://www.w3.org/2002/07/owl#Class, but not https://d-nb.info/standards/elementset/gnd#SubjectHeading.  I'm not too worried about this though because we've violated it (see our use of edm:), libraries violate it all the time, and the nature our the object of our triple is a "subject heading."
2. I'm more concerned with the latter. This property expects to be used by a rdfs:Class (our fedora object) that is a https://d-nb.info/standards/elementset/gnd#DifferentiatedPerson.  Technically, we have the same problem as above, but I don't think we have a "differentiated person" or a "person" conceptually.  Instead, we're describing a digital object. Using this in this way doesn't sit right with me.

Finally, I don't think we should go looking for an alternative here.  Let's just keep it simple and use the same properties we're already using here.

markpbaggett (changes requested):
I'm pretty good here, but I feel strongly that we shouldn't use `gnd:playedInstrument`.

docs/source/top_level_elements/genre.rst
CanOfBees: @markpbaggett thanks for your thoughts here. In the process of reviewing a different section of the mapping document, I realized we should be sticking with `opaquenamespace` for these. I will update this shortly.

docs/source/top_level_elements/genre.rst
```diff
@@ -0,0 +1,163 @@
...
+
+    <genre authority="dct">text</genre>
+    <genre>letter</genre>
+
+.. code-block:: turtle
+
+    @prefix edm: <http://www.europeana.eu/schemas/edm/> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:type "text" ;
```
mlhale7: Sorry I didn't catch this previously, but is it possible for us to use URIs for genre[@authority="dct"]/dcterms:type? Islandora has outlined the values in their mapping. Also, technically the capitalization is different in dcterms versus MODS typeOfResource, so "text" in MODS is "Text" in dcterms. Ultimately though it would be best if we could write in a URI for dcterms. This would make it consistent with the other values from MODS typeOfResource.

docs/source/top_level_elements/genre.rst
```diff
@@ -0,0 +1,163 @@
...
+
+    @prefix edm: <http://www.europeana.eu/schemas/edm/> .
+
+    <https://example.org/object/1> edm:hasType <http://id.loc.gov/authorities/subjects/sh85040974> .
+
+and `archivision:1754 <https://digital.lib.utk.edu/collections/islandora/object/archivision:1754/datastream/MODS/view>`_.
+
+.. code-block:: xml
+
+    <genre authority="lcsh" valueURI="http://id.loc.gov/authorities/subjects/sh85139020">Twentieth century</genre>
```
mlhale7, line 87: Values like these might ideally be placed in a temporal subject, but I don't think there is an easy way for us to separate these out (aka - problematic, but does not need to be addressed)

docs/source/top_level_elements/genre.rst
```diff
@@ -0,0 +1,163 @@
...
+        dcterms:subject <http://id.loc.gov/authorities/performanceMediums/mp2013015074> ;
+        dcterms:subject <http://id.loc.gov/authorities/performanceMediums/mp2013015342> ;
+        dcterms:subject <http://id.loc.gov/authorities/performanceMediums/mp2013015748> ;
+        dcterms:subject <http://id.loc.gov/authorities/performanceMediums/mp2013015540> ;
+        dcterms:type <http://id.loc.gov/authorities/genreForms/gf2014027156> ;
+        dcterms:type <http://id.loc.gov/authorities/genreForms/gf2014026956> ;
+        dcterms:type <http://id.loc.gov/authorities/genreForms/gf2014026097> ;
+        dcterms:type <http://id.loc.gov/authorities/subjects/sh99001779> .
+
+genre[@authority = 'lctgm']
```
mlhale7, line 153: I removed this category of items by completing [https://jirautk.atlassian.net/browse/DIGITAL-833](https://jirautk.atlassian.net/browse/DIGITAL-833). This can be removed from the mapping.

mlhale7 (changes requested):

markpbaggett (approved):
"Sure, why not?"

"I understand."

"Okay. :ok_hand:"

"Thanks for coming." 

"Okay. :ok_hand:"

"Sure, why not?"

"I understand."

"I understand."

"Sounds good."

"Okay. :ok_hand:"

"Okay. :ok_hand:"

"Thanks for coming." 

"I understand."

"Okay. :ok_hand:"

"I understand."

"Sure, why not?"

CanOfBees:
cinco-face-time-party-snoozer

docs/source/top_level_elements/genre.rst
CanOfBees: @mlhale7 thanks for pointing this out. I think we may have talked about this and I overlooked making this change.

CanOfBees: @mlhale7 I updated this in d77bff1a - let me know if it doesn't resolve this comment.

docs/source/top_level_elements/genre.rst
```diff
@@ -9,7 +9,19 @@ This section describes the different values for `genre` that we currently have i
...
+| MODS genre[@authority='dct'] String | RDF Predicate | URI                                              | dcterms String |
++-------------------------------------+---------------+--------------------------------------------------+----------------+
+| text                                | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/txt> | Text           |
++-------------------------------------+---------------+--------------------------------------------------+----------------+
+| still image                         | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/img> | StillImage     |
++-------------------------------------+---------------+--------------------------------------------------+----------------+
+| image                               | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/img> | Image          |
++-------------------------------------+---------------+--------------------------------------------------+----------------+
+| collection                          | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/col> | Collection     |
```
mlhale7: @CanOfBees - just checking, does the genre element ever have a value of "collection"? If so, it's likely inaccurate. If not, I wonder if we should remove it from the table in this section.

CanOfBees: @mlhale7 It appears on a compound obj in Vol Voices: 

```
volvoices:12079
volvoices:12078
volvoices:12077
volvoices:12076
```

mlhale7: Thanks. The record does seem to attempt to describe all of the three items shared in the compound object, so we can let it go as a collection I guess. This would be problematic for PrimoVE as I'm currently highlighting these items as a "Digital Collection".

mlhale7: As we discussed in our meeting, we'll go ahead and make a ticket to remove this since it's atypical.

docs/source/top_level_elements/genre.rst
mlhale7: Thanks - this is great!

docs/source/top_level_elements/genre.rst
CanOfBees: @mlhale7 this has been updated. Thanks!

markpbaggett (approved):
:ok_hand:

docs/source/top_level_elements/genre.rst
CanOfBees: :+1:  I've updated the PR to reflect this.

mathewjordan (approved):
:cow: :bell: 👍 
```
@prefix dcterms: <http://purl.org/dc/terms/> .

<https://example.org/object/1>
    dcterms:subject <http://id.loc.gov/authorities/performanceMediums/mp2013015187> .
```

mlhale7 (approved):

PR #38, opened 2020-10-23
Add name to draft mapping.

markpbaggett:
What Does this Do?
==================

This adds our name decisions to draft mapping. I've also added a lookup table for MARC relators.

How Should This Be Tested?
==========================

1. Check that all tests pass.
2. Does the resulting decisions reflect our decisions.

Additional Notes
================

I combined some decisions but attempted to do it in a way that makes things more concise and easier to understand.  If I failed to accomplish this, say so.


docs/source/contents/4_draft_mapping.rst
```diff
@@ -668,6 +670,195 @@ All values within <tableOfContents> will be mapped to RDF in the same way. Below
...
+.. code-block:: turtle
+    :caption: Resulting turtle for Name missing a valueURI
+    :name: Resulting turtle for Name missing a valueURI
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators/> .
+
+    <https://example.org/objects/1>
+        relators:cre "Daniel, Charles R. (Charlie), Jr., 1930-" .
+
+If there is a :code:`mods:name/valueURI` but it's empty, use the string literal instead:
```
mlhale7, line 761: Sad that this exists, but thank you for accounting for this.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -668,6 +670,195 @@ All values within <tableOfContents> will be mapped to RDF in the same way. Below
...
+.. code-block:: turtle
+    :caption: Resulting turtle from `volvoices:2495 <https://digital.lib.utk.edu/collections/islandora/object/volvoices:2495/datastream/MODS>`_
+    :name: Resulting turtle from `volvoices:2495 <https://digital.lib.utk.edu/collections/islandora/object/volvoices:2495/datastream/MODS>`_
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators/> .
+
+    <https://example.org/objects/1>
+        relators:asn "Bemis Bro. Bag Company" .
+
+Keep all Roles
```
mlhale7: Can we rename this use case to describe the situation better? Maybe just "Names with multiple roleTerms".

docs/source/contents/4_draft_mapping.rst
```diff
@@ -668,6 +670,195 @@ All values within <tableOfContents> will be mapped to RDF in the same way. Below
...
+
+Xpaths
+^^^^^^
+:code:`count(mods:name/mods:role)>1`
+
+Decision
+^^^^^^^^
+
+.. code-block:: xml
+    :caption: `harp:1 XML record <https://digital.lib.utk.edu/collections/islandora/object/harp%3A1/datastream/MODS>`_ has multiple role terms
```
mlhale7: In the previous example for roleTerms you used this same record. Since this record has more than one roleTerm, I don't think it's appropriate to use in both the previous example and this example. Can you pick another record that has just one roleTerm for it? We also might want to slightly rename the use case so that it's clear that we're dealing with names that have a single roleTerm. "Leverage Marc Relators for RDF Property Value and Relationship to the Digital Object" doesn't tell us anything about the number. In a way this could be included as an additional example in the previous bucket if you'd rather do that as well.

mlhale7:
@markpbaggett - one thing that we talked about briefly but didn't necessarily decide anything firm on is whether or not we want to migrate name values of "Unknown" (or "unknown" etc.). I'm assuming we'll go ahead with migrating it (which I'm fine with), since it hasn't been singled out to be dropped in this PR.

docs/source/contents/4_draft_mapping.rst
markpbaggett: :+1:

markpbaggett:
> @markpbaggett - one thing that we talked about briefly but didn't necessarily decide anything firm on is whether or not we want to migrate name values of "Unknown" (or "unknown" etc.). I'm assuming we'll go ahead with migrating it (which I'm fine with), since it hasn't been singled out to be dropped in this PR.

IMHO:  migrate, but just treat as string.

mlhale7:
> > @markpbaggett - one thing that we talked about briefly but didn't necessarily decide anything firm on is whether or not we want to migrate name values of "Unknown" (or "unknown" etc.). I'm assuming we'll go ahead with migrating it (which I'm fine with), since it hasn't been singled out to be dropped in this PR.
> 
> IMHO: migrate, but just treat as string.

That works for me - wasn't sure if we had officially decided.

markpbaggett:
@mlhale7 I think those commits should address your feedback.  Let me know if I missed something.

mlhale7 (approved):
@markpbaggett - thanks for the changes.

CanOfBees (approved):
Approved

mathewjordan (approved):
I appreciate the appendices and envy your ability to be concise.  :monocle_face:

PR #39, opened 2020-10-30
originInfo mapping

CanOfBees:
**JIRA Issue**: `DIGITAL-754 <https://jirautk.atlassian.net/browse/DIGITAL-754>`_

What Does this Do?
==================
This builds upon the originInfo analysis and provides final decisions to mapping document.

How Should This Be Tested?
==========================
Do tests pass? (probably not!...)
Are the contents consistent and properties justifiable?
Are slepping orrers verywhere?

Additional Notes
================
Things should hopefully be faithful to the analysis document.


docs/source/contents/4_draft_mapping.rst
```diff
@@ -862,6 +862,298 @@ Do not migrate.
...
+:code:`originInfo/dateOther[@encoding='edtf']` OR
+:code:`originInfo/dateOther[@encoding='edtf'][@point='end']` OR
+:code:`originInfo/dateOther[@encoding='edtf'][@point='start']`
+
+Decisions
+^^^^^^^^^
+
+As part of leveraging the EDTF format, some conversion will be necessary; e.g. translating date strings to EDTF values as in the following example. The `dcterms:date` property was selected.
+
+`kintner:56 <https://digital.lib.utk.edu/collections/islandora/object/kintner%3A56>`_
```
mlhale7: This entire collection (kintner) was remediated this summer, so the example is no longer relevant. Can you select another instance of dateOther (assuming there are still several hanging out) from another collection? As a group, we should talk more about what are expectations are for our mapping matching our present metadata since the metadata is always changing.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -862,6 +862,298 @@ Do not migrate.
...
+
+Decision
+^^^^^^^^
+
+The majority of the applicable values are associate with a `@valueURI`.  The `relators:pup` property was selected.
+
+.. code-block:: xml
+
+    <originInfo>
+        <place supplied="yes">
```
mlhale7, line 1103: It looks like we are disregarding the supplied attribute (which I am fine with). Can we specify that we are making this decision?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -862,6 +862,298 @@ Do not migrate.
...
+
+Justification
+^^^^^^^^^^^^^
+
+No dispute on the values contained in `publisher`.
+
+XPath
+^^^^^
+
+:code:`originInfo/publisher`
```
mlhale7: I didn't notice this until this Xpath, but you are using a less verbose Xpath throughout. @markpbaggett and @mathewjordan, for consistency do we always want to include "mods" - aka mods:originInfo/mods:publisher?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -862,6 +862,298 @@ Do not migrate.
...
+.. code-block:: turtle
+
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
+
+    <https://example.org/objects/1> relators:pbl "Frederick D. Benteen" ;
+        relators:pup <http://id.loc.gov/authorities/names/n79006530> .
+
+originInfo/issuance
+-------------------
+We will not be migrating `issuance`.
```
mlhale7, line 1184: We might want to elaborate a little on why here - it isn't consistently applied, we don't currently display this value publicly, etc.

mlhale7: I'd also include all the sections - Use Case, Justification, Xpath, Decision.

mlhale7 (changes requested):
Final question - didn't we have copyrightDate ([calahan:1](https://digital.lib.utk.edu/collections/islandora/object/calahan%3A1)) and maybe even dateCaptured in the initial analysis doc?

docs/source/contents/4_draft_mapping.rst
CanOfBees: @mlhale7 thanks for catching this - I've updated w/a playbills example.

docs/source/contents/4_draft_mapping.rst
CanOfBees: @mlhale7 - I think I should follow suit and add the namespace prefix.

docs/source/contents/4_draft_mapping.rst
CanOfBees: @mlhale7 can you clarify? Would you prefer `` :code:`originInfo/place[\@supplied='yes']/placeTerm[\@text][\@valueURI]` `` in the XPath section? Or am I completely missing your point? 

Thanks!

mlhale7: I was wondering if we wanted to acknowledge the loss of this information in the document (though as I said, I don't think much is lost). No Xpath changes are necessary. I'm happy to approve with this, it just wasn't something we discussed.

docs/source/contents/4_draft_mapping.rst
mlhale7: As we discussed in our meeting, we'll deal with consistency in the final editing. A strong argument was made for being less verbose.

docs/source/contents/4_draft_mapping.rst
mlhale7: @CanOfBees - thanks for the additions.

mlhale7 (approved):

markpbaggett (approved):
Sounds good.

PR #40, opened 2020-11-03
Subjects

mlhale7:
**JIRA Issue**: [DIGITAL-599](https://jirautk.atlassian.net/browse/DIGITAL-599)

What Does this Do?
==================

This PR adds analysis for the top-level element "subject" and all of its child elements.

How Should This Be Tested?
==========================

Let me know if anything is missing from the analysis that needs to be considered. Check to see if the properties used are appropriate. Provide suggestions on any improvements to help with legibility and consistency.

Additional Notes
================

When I add this to the mapping, I likely will need to collapse this document more into fewer categories. I'm getting this out for review as it has been hanging out far too long.

### DIFF
--- a/docs/source/top_level_elements/subject.rst
+++ b/docs/source/top_level_elements/subject.rst
@@ -207,7 +207,7 @@ If only a string is available, like with , we will use that:
 Temporal subjects
 -----------------
 
-None of our existing temporal subjects include URIs, so all will be treated as strings. These values are prominent in
+None of our existing temporal subjects include URIs. Most will be treated as strings. These values are prominent in
 Volunteer Voices and the Pi Beta Phi to Arrowmont collections.
 
 `Example of temporal subject - volvoices:1833 <https://digital.lib.utk.edu/collections/islandora/object/volvoices%3A1833/datastream/MODS/view>`_.
@@ -224,6 +224,21 @@ Volunteer Voices and the Pi Beta Phi to Arrowmont collections.
     @prefix schema: <http://schema.org/> .
 
     <https://example.org/objects/1> schema:temporalCoverage "Era 9 - Postwar United States (1945-1970's)" .
+
+In addition to these textual values, UTK does have temporal subjects that share numeric dates in EDTF format. These are
+primarily from the Volunteer Voices collection. `Here's an example record - volvoices:2945 <https://digital.lib.utk.edu/collections/islandora/object/volvoices%3A2945/datastream/MODS/view>`_.
+
+.. code-block:: xml
+
+    <subject>
+        <temporal>1970-09-30</temporal>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix schema: <http://schema.org/> .
+
+    <https://example.org/objects/1> schema:temporalCoverage "1970-09-30" .
 
 Geographic subjects
 -------------------
--- a/docs/source/top_level_elements/subject.rst
+++ b/docs/source/top_level_elements/subject.rst
@@ -234,17 +234,117 @@ URIs will be used when present, but strings can be used when there is no URI.
     mods:subject[@valueURI]/mods:geographic
     mods:subject/mods:geographic[@valueURI]
 
-    Adding examples....
+`Here's an example where the URI is present on the subject - webster:1127 <https://digital.lib.utk.edu/collections/islandora/object/webster%3A1127/datastream/MODS/view>`_.
+
+.. code-block:: xml
+
+    <subject authority="geonames" valueURI="http://sws.geonames.org/4050810">
+        <geographic>The Sawteeth</geographic>
+        <cartographics>
+            <coordinates>35.64342, -83.36237</coordinates>
+        </cartographics>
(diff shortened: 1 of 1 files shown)


docs/source/top_level_elements/subject.rst
```diff
@@ -0,0 +1,497 @@
...
+        dcterms:subject "Pollution" ;
+        dcterms:subject "Knight" .
+
+Topical subjects with URIs
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Note the variations in where @valueURI is placed in the Xpaths listed below.
+
+    mods:subject[@valueURI]/mods:topic
+    mods:subject/mods:topic[@valueURI]
```
markpbaggett: I don't care that these are not treated as inline code, but I'm pointing it out in case you do.

docs/source/top_level_elements/subject.rst
```diff
@@ -0,0 +1,497 @@
...
+    <subject>
+        <geographic>Covington (Tenn.)</geographic>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:spatial "Covington (Tenn.)" .
+
```
markpbaggett, line 299: Just pointing out about spatial.  I don't care a ton, but I felt like I should at least mention. This property has a http://purl.org/dc/dcam/rangeIncludes (you can think of this as something like an rdfs:range / literally defined as "A suggested class for values of this property.") of http://purl.org/dc/terms/Location.

First, because it has a rangeIncludes it expects the value to be an instance of that class.  In this case, http://purl.org/dc/terms/Location. Obviously we have a string literal which is for sure not an instance of this class.  We've dealt with this a lot (see our use of edm) so I don't feel strongly, but in all those case we've minimally had a URI even if the URI was not an instance of what the range like property expected. A location is defined as "A spatial region or named place." We have that so perhaps we don't care.

docs/source/top_level_elements/subject.rst
```diff
@@ -0,0 +1,497 @@
...
+    </subject>
+    <subject>
+        <topic valueURI="http://id.loc.gov/authorities/subjects/sh85023396">Child welfare</topic>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "http://id.loc.gov/authorities/subjects/sh85023396" .
```
markpbaggett, line 136: I don't care, but just pointing out that we only model the second xml is example as :turtle: .  It doesn't matter in my opinion, but thought I'd add.

docs/source/top_level_elements/subject.rst
```diff
@@ -0,0 +1,497 @@
...
+        </cartographics>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:spatial <https://sws.geonames.org/4630912> ;
+        dcterms:spatial "35.58546, -83.84444" .
+
```
markpbaggett, line 357: Same note as above.  Again don't care.

docs/source/top_level_elements/subject.rst
```diff
@@ -0,0 +1,497 @@
...
+        </cartographics>
+    </subject>
+
+All that is needed in this case is to bring over the URI.
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:spatial <https://sws.geonames.org/4630912> .
```
markpbaggett, line 337: Same note as above.  Again don't care.

markpbaggett (commented):
@mlhale7, I have comments but not necessarily requested changes.  Read through them and make sure you are okay and if so just say so and we can merge as is.

CanOfBees (approved):
@mlhale7 - thanks for your work. I think this looks good, and only have a general comment: 

building off of @markpbaggett's comments regarding `dcterms:spatial` and location, would it make any sense to have work on a partial list of URIs for migration (or post-migration cleanup)? E.g. we approximate which of the strings are places, and then identify URIs for those?

I'm not advocating that we do this as part of migration - maybe this is more a  post-migration cleanup thing?

In any case, thanks!

docs/source/top_level_elements/subject.rst
mlhale7: Yes, this way me being lazy. I have just committed this change.

docs/source/top_level_elements/subject.rst
mlhale7: Yes, since the other value is geographic, I thought it might be confusing to include it in the turtle. Probably it would've been best to simply not include the geographic subject in the given XML. I'll make that change.

markpbaggett:
Switching to approve.

markpbaggett (approved):

PR #41, opened 2020-11-10
Genre mapping

CanOfBees:
**JIRA Issue**: [DIGITAL-584](https://jirautk.atlassian.net/browse/DIGITAL-584)

What Does this Do?
==================
Adds a `genre` mapping to our draft mapping document.

How Should This Be Tested?
==========================
Do all tests pass?
Any glaring spelling|punctuation|grammatical errors?

Additional Notes
================
I struggled to remember and/or articulate justifications for several of the decisions made for `genre`. Feedback in the form of "You dolt! We discussed __this__ on __some date__. Argh!" would be deeply appreciated.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -1794,8 +1794,8 @@ genre
 | edm:hasType     | URI/String Literal | N/A   |
 +-----------------+--------------------+-------+
 
-genre: values that map to dcterms:type and edm:hasType
-------------------------------------------------------
+genre: values that map to dcterms:type
+--------------------------------------
 
 Use Case
 ^^^^^^^^
@@ -1818,8 +1818,6 @@ For values *outside* of the following table, we selected the `edm:hasType` prope
 +-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
 | text                                          | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/txt> | Text               |
 +-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
-| *                                             | edm:hasType   | <http://www.europeana.eu/schemas/edm/>           | N/A                |
-+-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
 
 XPaths
 ^^^^^^
@@ -1827,15 +1825,14 @@ XPaths
 :code:`genre[not(@*)][string() = 'notated music']` OR
 :code:`genre[@authority = 'dct'][string() = 'image']` OR
 :code:`genre[@authority = 'dct'][string() = 'still image']` OR
-:code:`genre[@authority = 'dct'][string() = 'text']` OR
-:code:`genre[not(@*) and not(string() = ('cartographic','notated music'))]`
+:code:`genre[@authority = 'dct'][string() = 'text']`
 
-Alternately, the first set of XPaths can be notated as:
+Alternately, these XPaths can be notated as:
 :code:`genre[(not(@*) and (string() = ('cartographic', 'notated music')) or (@authority = 'dct' and (string() = ('text', 'image', 'still image')))]`
 
 Decision
 ^^^^^^^^
-The property `edm:hasType` was selected for values that fall outside of the preceding table. For values that match, `dcterms:type` has been selected.
+The `dcterms:type` property has been selected.
 
 `Example record - volvoices:11551 <https://digital.lib.utk.edu/collections/islandora/object/volvoices:11551/datastream/MODS/view>`_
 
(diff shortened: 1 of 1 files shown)


docs/source/contents/4_draft_mapping.rst
```diff
@@ -1784,6 +1784,263 @@ subject
...
+    <genre>notated music</genre>
+    <genre authority="dct">still image</genre>
+    <genre>sheet music</genre>
+
+.. code-block:: turtle
+
+    @prefix edm: <http://www.europeana.eu/schemas/edm/> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> edm:hasType "notated music", "sheet music" ;
```
mlhale7: I hadn't realized that we had MODS typeOfResource values in both genre and genre[@authority="dct"]. What are your thoughts on looking up a URI based on the strings "text", "still image", and "image"? (sort of how was done for roleTerms in the name analysis) Otherwise we are going to have strings and URIs for the same value. This is also potentially something we can fix later on.

markpbaggett (approved):
Sounds good

docs/source/contents/4_draft_mapping.rst
CanOfBees: @mlhale7 I updated this to what I think we're after, but please let me know if you'd like to see it changed.

Thanks!

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1784,6 +1784,280 @@ subject
...
+| notated music                                 | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/not> | Notated music      |
++-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
+| still image                                   | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/img> | Still image        |
++-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
+| text                                          | dcterms:type  | <http://id.loc.gov/vocabulary/resourceTypes/txt> | Text               |
++-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
+
+XPath
+^^^^^
+:code:`genre[not(@*)]`
```
mlhale7: @CanOfBees - the table looks good. For your Xpath I wondered if we needed to be more specific. You've sort of combined two categories into one. Thinking about this Xpath being machine actionable, how would we distinguish between the genre values that go into edm:hasType and those in dcterms:type? You've made this explicit in your Decision and Justification text, but the Xpath encompasses both. I'm wondering if splitting this into two cases where in one you say the Xpath is genre[string()="text"] etc. and in the other the Xpath is genre[not(string()="text")] would clarify this.

CanOfBees: @mlhale7 I tried to take a stab at merging the two sections, as well as updating the table and xpaths. I don't know if I'm happy with the results. :man_shrugging: Please let me know what you think about the changes - do they help clarify at all, or have I just made things worse?

Thanks!

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1794,16 +1794,16 @@ genre
 | edm:hasType     | URI/String Literal | N/A   |
 +-----------------+--------------------+-------+
 
-genre[not(@*)]
---------------
+genre[not(@*)] AND genre[@authority='dct']
```
mlhale7: Perhaps slightly rename this (since we aren't including ALL genre[not(@*)] values here)? Maybe something less Xpath focused like "Values that map to dcterms:type values"?

CanOfBees: Thank you!

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1784,6 +1784,271 @@ subject
...
++-----------------------------------------------+---------------+--------------------------------------------------+--------------------+
+
+XPaths
+^^^^^^
+:code:`genre[not(@*)][string() = 'cartographic']` OR
+:code:`genre[not(@*)][string() = 'notated music']` OR
+:code:`genre[@authority = 'dct'][string() = 'image']` OR
+:code:`genre[@authority = 'dct'][string() = 'still image']` OR
+:code:`genre[@authority = 'dct'][string() = 'text']` OR
+:code:`genre[not(@*) and not(string() = ('cartographic','notated music'))]`
```
mlhale7: I guess I didn't notice this one on the end. These are the edm:hasType values. Is there any reason why we can't just make this a separate category instead of putting it in with dcterms:type? I assume that everything in the same category is treated the same way as everything else in the category. Maybe this is a faulty assumption on my part, but I think it makes interpreting this doc easier.

mlhale7 (approved):
@CanOfBees - thanks for these changes. This looks good to me.

mlhale7 (approved):

markpbaggett (approved):
Looks good to me too! :+1:

PR #42, opened 2020-11-17
Add a sample abstract doc.

markpbaggett:
What Does this Do?
==================

Adds a sample abstract top level element doc.

How Should This Be Tested?
==========================

Do tests pass? Do you see anything wrong with my description of the document.

Additional Notes
================

I wasn't sure if I was supposed to open directly against draft mapping, so I wrote it so that it could be easily applied.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -629,7 +629,7 @@ Regardless of the number, the value has the same semantic relationship to the ob
 Xpath
 ^^^^^
 
-:code:`mods:abstract[text()!=""]`
+:code:`mods:abstract[text()]`
 
 Decision
 ^^^^^^^^
@@ -682,7 +682,7 @@ The value of the text node has no semantic meaning or value.
 Xpaths
 ^^^^^^
 
-:code:`mods:abstract[text()=""]`
+:code:`mods:abstract[string()=""]`
 
 Decision
 ^^^^^^^^
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -606,6 +606,89 @@ We will not retain data regarding displayLabel attributes moving forward.
 abstract
 ========
 
++------------------+------------+-------------------+-----------------------------------------------------+
+| Predicate        | Value Type | Range (if needed) | Usage notes                                         |
++==================+============+===================+=====================================================+
+| dcterms:abstract | Literal    |                   | Use for all mods:abstracts that are not blank nodes |
++------------------+------------+-------------------+-----------------------------------------------------+
+
+Abstracts that are not Blank Nodes
+----------------------------------
+
+Use Case
+^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


docs/source/top_level_elements/abstract.rst
```diff
@@ -0,0 +1,85 @@
...
+we will still map to dc:terms abstract.
+
+.. code-block:: xml
+
+    <abstract>
+        Postcard with handwritten note sent from Knoxville to Miss Virginia Bogart, Loudon, Tennessee on March 2, 1944 for a postage of 1 cent.
+    </abstract>
+    <abstract>
+        The hardwood forest of America, and probably of the entire world, originated in the Great Smoky Mountains, where remains the nation's largest body of virgin hardwood forest, and the world's greatest variety of trees, flowering shrubs and wild flowers.
+    </abstract>
```
mlhale7, line 56: What are people's thoughts on concatenating the two (or more) strings values together in the metadata? I know we end up doing this for display. Typically when I remediate I do this, or move the second value to a note instead. In this particular case we have a direct transcription from the source. If I were doing this manually, I might add "Postcard reads 'quote'" and move it to a note. In terms of consistency, in the vanvactor collection we have decided to treat transcribed handwriting as a basic note undistinguished from others. Keeping them separate would make it easier to move things later perhaps.

mlhale7 (commented):
@markpbaggett - I think it would be fine for us to commit this directly to the mapping. There aren't that many things to consider since we're dealing with strings and the property to use is pretty straightforward. Overall it will save us time.

docs/source/top_level_elements/abstract.rst
markpbaggett: I hate if from the point of view of technical burden.  It's much more straightforward if we just use repeating nodes.

markpbaggett:
> @markpbaggett - I think it would be fine for us to commit this directly to the mapping. There aren't that many things to consider since we're dealing with strings and the property to use is pretty straightforward. Overall it will save us time.

done

docs/source/contents/4_draft_mapping.rst
```diff
@@ -606,6 +606,89 @@ We will not retain data regarding displayLabel attributes moving forward.
...
+
+Justification
+^^^^^^^^^^^^^
+
+The value of the text node has no semantic meaning or value.
+
+Xpaths
+^^^^^^
+
+:code:`mods:abstract[text()=""]`
```
CanOfBees: Let's make this `abstract[string() = '']` :)

CanOfBees (changes requested):

docs/source/top_level_elements/abstract.rst
mlhale7: I'd agree that this isn't really practical to think about combining select values. I'm in agreement with using repeated nodes, but wanted to bring this up for discussion.

mathewjordan (approved):
🧑‍🎨

mlhale7 (approved):
@markpbaggett - thanks for your work on this.

CanOfBees (approved):
I understand.

PR #43, opened 2020-11-17
Subjects mapping

mlhale7:
**JIRA Issue**: [DIGITAL-599](https://jirautk.atlassian.net/browse/DIGITAL-599)

What Does this Do?
==================

This adds decisions for the subject top-level element to the mapping doc and streamlines some of the categories from the analysis doc.

How Should This Be Tested?
==========================

Does the automated testing pass? Are the Xpaths defined actionable and distinct from one another?

Additional Notes
================

Thanks for reviewing!

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -1799,9 +1799,9 @@ These nodes contain no information.
 Xpath
 ^^^^^
 
-:code:`subject/topic=''` OR
-:code:`subject/geographic=''` OR
-:code:`subject/name/namePart=''`
+    :code:`subject/topic[text()='']` OR
+    :code:`subject/geographic[text()='']` OR
+    :code:`subject/name/namePart[text()='']`
 
 Decision
 ^^^^^^^^
@@ -1846,10 +1846,10 @@ Xpath
 
 Note that there is inconsistency in where the valueURI attribute is placed.
 
-:code:`mods:subject[@valueURI]/mods:topic` OR
-:code:`mods:subject/mods:topic[@valueURI]` OR
-:code:`mods:subject[@valueURI]/mods:name/mods:namePart` OR
-:code:`mods:subject/mods:name[@valueURI]/mods:namePart`
+    :code:`mods:subject[@valueURI]/mods:topic` OR
+    :code:`mods:subject/mods:topic[@valueURI]` OR
+    :code:`mods:subject[@valueURI]/mods:name/mods:namePart` OR
+    :code:`mods:subject/mods:name[@valueURI]/mods:namePart`
 
 Decision
 ^^^^^^^^
@@ -1944,6 +1944,17 @@ with string values that are: Charlie Daniel Cartoon Collection, Ed Gamble Cartoo
 North America Records, the American Civil War Collection, Ramsey Family Papers, Tennessee Documentary History,
 and Volunteer Voices.
 
+The Volunteer Voices collection includes subjects with three different displayLabel values - Volunteer Voices Curriculum Topics,
+Tennessee Social Studies K-12 Eras in American History, and Broad Topics. These subjects are currently given separate
+facets in Islandora's metadata display. Discovery to the collection via two of these subject categories is also featured
+on the `Tennessee State Library and Archives website <https://sos.tn.gov/products/tsla/volunteer-voices>`_ (Broad Topics
+and Tennessee Social Studies K-12 Eras in American History). While these subjects have been distinguished previously from
+other subjects in the past by their distinct Xpath, having so many different types of subjects was found to be unnecessary
+going forward. Broad Topics and Curriculum Topics will be folded in with all other subjects. For links to external websites,
+like TSLA's, we can use the string values to supply a link without needing to place them in a separate property. Note that
+subjects associated with Tennessee Social Studies K-12 Eras in American History are dealt with
+separately below.
+
 Justification
 ^^^^^^^^^^^^^
 
@@ -1953,8 +1964,8 @@ standpoint, strings still support discovery.
 Xpath
 ^^^^^
 
-:code:`mods/subject[not(@valueURI)]/topic[not(@valueURI)]` OR
-:code:`mods/subject[not(@valueURI)]/name[not(valueURI)]/namePart[not(valueURI)]`
+    :code:`mods/subject[not(@valueURI)]/topic[not(@valueURI)]` OR
+    :code:`mods/subject[not(@valueURI)]/name[not(valueURI)]/namePart[not(valueURI)]` OR
 
 Decision
 ^^^^^^^^
@@ -2003,6 +2014,34 @@ Here's an `example where only a string value is available for a name - gamble:14
 
     <https://example.org/objects/1> dcterms:subject "Xerox Corporation" .
 
+Here's an `example from Volunteer Voices of a "Broad Topics" subject - volvoices:4058 <https://digital.lib.utk.edu/collections/islandora/object/volvoices%3A4058/datastream/MODS/view>`_.
+
+.. code-block:: xml
+
+    <subject displayLabel="Broad Topics">
+        <topic>Frontier Settlement and Migration</topic>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "Frontier Settlement and Migration" .
+
+Here's an `example of @displayLabel="Volunteer Voices Curriculum Topics" - volvoices:2141 <https://digital.lib.utk.edu/collections/islandora/object/volvoices%3A2141/datastream/MODS/view>`_.
+
+.. code-block:: xml
+
+    <subject displayLabel="Volunteer Voices Curriculum Topics">
+        <topic>Civil Rights movement in Tennessee</topic>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "Civil Rights movement in Tennessee" .
+
 Temporal subjects
 -----------------
 
@@ -2022,13 +2061,14 @@ retaining these values is particularly important.
 Xpath
 ^^^^^
 
-:code:`mods/subject[not(@displayLabel)]/temporal`
+:code:`mods/subject/temporal`
 
 Decision
 ^^^^^^^^
 
 Temporal subjects without the displayLabel attribute will be directly mapped as strings to schema:temporalCoverage. This
-property was chosen because it allows a wider range of values than other potential solutions (such as ).
+property was chosen because it allows a wider range of values than other potential solutions (such as dcterms:temporalCoverage
+which requires that the value is part of the class PeriodOfTime).
 
 `Example of temporal subject - arrow:268 <https://digital.lib.utk.edu/collections/islandora/object/arrow%3A268>`_.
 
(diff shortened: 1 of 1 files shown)


docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+^^^^^^^^^^^^^
+
+These nodes contain no information.
+
+Xpath
+^^^^^
+
+    :code:`subject/topic[text()='']` OR
+    :code:`subject/geographic[text()='']` OR
+    :code:`subject/name/namePart[text()='']`
```
CanOfBees: I know that I'm being a bit ridiculous, but this syntax isn't right. Let's change it; all of the following versions are better:
`subject/topic[not(text())]` or
`subject/topic[string() = '']` or
`subject/topic[. = '']`

The second one is probably the best and/or safest.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+            <roleTerm authority="marcrelator" valueURI="http://id.loc.gov/vocabulary/relators/pht">Photographer</roleTerm>
+        </role>
+    </name>
+    <title>Aunt Lydia's cave by the creek, Francisa.</title>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "Whaley, Aunt Lydia"
```
CanOfBees: I know that you know this, but we're missing a `.` here at the end of 2416.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "http://id.loc.gov/authorities/subjects/sh85147554" ;
+        dcterms:subject "http://id.loc.gov/authorities/subjects/sh85147447" ;
+        dcterms:subject "http://id.loc.gov/vocabulary/graphicMaterials/tgm008085" .
+
+`cdf:5384 as an example of mods:subject
+/mods:topic[@valueURI] <https://digital.lib.utk.edu/collections/islandora/object/cdf%3A5384/datastream/MODS/view>`_
```
CanOfBees: Looks like an extra Return/linefeed between line 1882 and 1883 here?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+        <topic>Portraits</topic>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "http://id.loc.gov/authorities/subjects/sh85147554" ;
+        dcterms:subject "http://id.loc.gov/authorities/subjects/sh85147447" ;
+        dcterms:subject "http://id.loc.gov/vocabulary/graphicMaterials/tgm008085" .
```
CanOfBees: I'm probably forgetting something from one of our meetings, but shouldn't our URIs be wrapped in angle brackets; e.g. `<http://id.loc.gov/authorities/subjects/sh85147554>`?

Sorry if I'm blanking on a previously discussed topic.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+
+    <subject>
+        <topic valueURI="http://id.loc.gov/authorities/subjects/sh85023396">Child welfare</topic>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "http://id.loc.gov/authorities/subjects/sh85023396" .
```
CanOfBees: Same question regarding double quotes vs angle brackets.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+        <name>
+            <namePart>United States. Army. Medical Corps</namePart>
+        </name>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "http://id.loc.gov/authorities/names/n85185770" .
```
CanOfBees: Same question regarding double quotes vs angle brackets.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+            <namePart>Arthur, Joseph Charles, 1850-1942</namePart>
+        </name>
+    </subject>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:subject "http://id.loc.gov/authorities/names/n88144876" ;
+        dcterms:subject "http://id.loc.gov/authorities/names/n87116131" .
```
CanOfBees: Same question regarding double quotes vs angle brackets.

CanOfBees (changes requested):
Interspersed some comments.

Overall, you make me want to rewrite all of my mapping material. Thanks for providing such good use case/justification sections - they're awesome.

docs/source/contents/4_draft_mapping.rst
markpbaggett: @CanOfBees is definitely right, but I'm wondering why `subject/topic[not(text())]` is not the best.  To me, this explicitly states that this should not match on subject topics where there is no text node.  Not saying @CanOfBees is wrong, but I'm curious why 2 is better than 1. @CanOfBees can you explain just so I better understand and appreciate more?

docs/source/contents/4_draft_mapping.rst
mlhale7: Yep, copy and paste from a previous text string got me. Thank you.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks @CanOfBees for noting it though - yesterday got away from me.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+Justification
+^^^^^^^^^^^^^
+
+These nodes contain no information.
+
+Xpath
+^^^^^
+
+    :code:`subject/topic[text()='']` OR
+    :code:`subject/geographic[text()='']` OR
```
markpbaggett: Similarly to @CanOfBees other comments, this should be one of these:

```
subject/geographic[not(text())] 
subject/geographic[string()='']
subject/geographic[.=''] 
```

docs/source/contents/4_draft_mapping.rst
markpbaggett: And this should be one of these:

```
subject/name/namePart[not(text())] 
subject/name/namePart[string() = '']
subject/name/namePart[. = ''] 
```

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+Here's an example from `roth - roth:1587 <https://digital.lib.utk.edu/collections/islandora/object/roth%3A1587/datastream/MODS/view>`_.
+
+.. code-block:: xml
+
+    <subject>
+        <name authority="" valueURI="">
+            <namePart/>
+            </name>
+    </subject>
+
```
markpbaggett, line 1850: You don't have a subjec/geographic example.  Did you want to have one?

docs/source/contents/4_draft_mapping.rst
mlhale7: Yes, I've apparently polluted half of my analysis with this! Addressing now.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks - I went with the second as suggested.

docs/source/contents/4_draft_mapping.rst
markpbaggett: :100: 

Good catch!  That would have been bad.  I wonder if we've been checking for this consistently.

docs/source/contents/4_draft_mapping.rst
markpbaggett: Again, :100:. :clap: 

We must check for this throughout this document.

docs/source/contents/4_draft_mapping.rst
markpbaggett: :clap:

docs/source/contents/4_draft_mapping.rst
markpbaggett: :100: :clap:

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,644 @@ Decision
...
+^^^^^^^^^^^^^
+
+Subject values are important access points for users that require migration. While URIs would be ideal from a technical
+standpoint, strings still support discovery.
+
+Xpath
+^^^^^
+
+    :code:`mods/subject[not(@valueURI)]/topic[not(@valueURI)]` OR
+    :code:`mods/subject[not(@valueURI)]/name[not(valueURI)]/namePart[not(valueURI)]` OR
```
markpbaggett: There is an extra OR, but I don't care.

docs/source/contents/4_draft_mapping.rst
markpbaggett: :+1:

markpbaggett (changes requested):

docs/source/contents/4_draft_mapping.rst
mlhale7: Thank you - I addressed this in the most recent commit.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thank you - I addressed this in the most recent commit.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thank you - I addressed this in the most recent commit.

docs/source/contents/4_draft_mapping.rst
markpbaggett: I'm good but in case you're curious from the definition of the string function on  W3C:

> Every node has a string value, even an element with element-only content (which has no typed value). Moreover, casting an atomic value to a string always succeeds. Functions, maps, and arrays have no string value, so these are the only arguments that satisfy the type signature but cause failure.

docs/source/contents/4_draft_mapping.rst
markpbaggett: This is a question, but I'm good if ya'll are good.

docs/source/contents/4_draft_mapping.rst
markpbaggett: Don't care, but in case you do, recommenting.

markpbaggett (approved):

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks for this. I think that we likely should address this for all top level elements instead of doing it individually for each one simply to decrease the sheer volume of this document.

I've gone ahead and added an example for subject/geographic so that all Xpaths listed are illustrated. If we decide to remove it later that's fine, but at least we'll have it fully documented here.

docs/source/contents/4_draft_mapping.rst
mlhale7: My initial instinct was subject/topic[not(text())], but I could not figure out which would be "safest". I would say from a consistency standpoint we do have [not(text())] elsewhere in our mapping. We should probably address this in the same way throughout just so that the doc is easier to digest.

docs/source/contents/4_draft_mapping.rst
mlhale7: I removed it - thank you for catching it.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1777,10 +1777,651 @@ Decision
 
 **Do not migrate!**
 
-
 subject
 =======
```
mathewjordan, line 1782: This is really silly for my to really only catch this, but looks like we need the predicate table for this still.

mlhale7: Thanks - I've just added a table.

markpbaggett (approved):
:+1:

mathewjordan (approved):
Yep, looks good. Thanks for adding the table!

CanOfBees (approved):
Thanks!

PR #45, opened 2020-11-20
Relateditem mapping

CanOfBees:
**JIRA Issue**: `DIGITAL-591 <https://jirautk.atlassian.net/browse/DIGITAL-591>`_

What Does this Do?
==================
An attempt at adding a mapping for relatedItem.

How Should This Be Reviewed?
==========================
As mentioned, I struggled to remember applicable details for many (or all) of the Justification sections. Any and all feedback and thoughts for improving these sections would be very appreciated!

Additional Notes
================

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -2152,21 +2152,23 @@ Drop all values in mods:part.
 relatedItem
 ===========
 
-relatedItem[not(@*)]
---------------------
+relatedItem as simple structural metadata
+-----------------------------------------
 Use Case
 ^^^^^^^^
-`relatedItem`, without attributes, is used in the Volunteer Voices collection. It expresses a structural relationship; that will be handled/expressed by default behavior in our next-gen DAMS.
+`relatedItem`, with and without attributes, is used in a number of collections to express structural relationships. These relationships will be handled/expressed by default behavior in our next-gen DAMS.
 
 Justification
 ^^^^^^^^^^^^^
 
 XPath
 ^^^^^
-:code:`relatedItem[not(@*)]`
+:code:`relatedItem[not(@*)]` OR
+:code:`relatedItem[@type='host'][@displayLabel[matches(., 'project') or matches(., 'Project') or matches(., 'Digital Collection') or matches(., 'Project Part')]
 
 Decision
 ^^^^^^^^
+Do not migrate.
 
 Hierarchical Sheet Music Identifier
 -----------------------------------
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -2311,16 +2311,77 @@ The `dcterms:bibliographicCitation` predicate was selected for these values.
     @prefix dcterms: <http://purl.org/dc/terms/> .
 
     <https://example.org/objects/1> dcterms:bibliographicCitation "The Arrow, Volume 27, Number 1" .
+
 relatedItem/identifier[@type]
 -----------------------------
 Use Case
 ^^^^^^^^
+This XPath's `@type` has three distinct values: `local`, `catalog`, and `pid`. `@type='pid'` is used in collection-level records to indicate featured items and should not be migrated. `@type='local'` is used to indicate the manuscript number of the resource's archival collection. `@type='catalog'` is used in the Van Vactor collection to indicate the identifying number for an alternate version of the score.
+
 Justification
 ^^^^^^^^^^^^^
 XPath
 ^^^^^
+:code:`relatedItem/identifier[@type]`
+
 Decision
 ^^^^^^^^
+`@type='local'`'s value, if present, maps in to the `dbo:collection` property.
+
+`Example record - heilman:26 <https://digital.lib.utk.edu/collections/islandora/object/heilman:261/datastream/MODS/view>`_
+
+.. code-block:: xml
+
+    <relatedItem type="host" displayLabel="Collection">
+      <titleInfo>
+        <title>Botany Department Photographs</title>
+      </titleInfo>
+      <identifier type="local">AR.0488</identifier>
+    </relatedItem>
+
+.. code-block:: turtle
+
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+
+    <https://example.org/objects/1> dbo:collection "Botany Department Photographs, AR.0488" .
+
+`@type='catalog'`'s value, if present, will be represented by the `opaque:sheetmusic_hostItem` property.
+
+`Example record - vanvactor:1 <https://digital.lib.utk.edu/collections/islandora/object/vanvactor:1/datastream/MODS/view>`_
+
+.. code-block:: xml
+
+    <relatedItem type="otherVersion">
+      <titleInfo>
+        <title>Gefunden</title>
+      </titleInfo>
+      <identifier type="catalog">M047</identifier>
+    </relatedItem>
+    <relatedItem displayLabel="Project" type="host">
+      <titleInfo>
+        <title>David Van Vactor Music Collection</title>
+      </titleInfo>
+    </relatedItem>
+    <relatedItem displayLabel="Collection" type="host">
+      <titleInfo>
+        <title>David Van Vactor Papers</title>
+      </titleInfo>
+      <identifier>MS.1942</identifier>
+      <location>
+        <url>https://n2t.net/ark:/87290/v8pz5703</url>
+      </location>
+    </relatedItem>
+
+.. code-block:: turtle
+
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+
+    <https://example.org/objects/1> dbo:collection "David Van Vactor Papers, MS.1942" ;
+        dbo:isPartOf <https://n2t.net/ark:/87290/v8pz5703> ;
+        opaque:sheetmusic_hostItem "Gefunden, M047" .
+
+`@type='pid'` will not be migrated.
 
 relatedItem/location[physicalLocation]
 --------------------------------------
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -2407,7 +2407,7 @@ Decision
 ^^^^^^^^
 See the following section on `relatedItem/identifier[@type]`.
 
-I'm not sure about the relevance of `relatedItem[@type='otherVersion']/location/url`. In theory, it would look something like this, but it would need to be added to our metadata/RDF after migration:
+The XPath `relatedItem[@type='otherVersion']/location/url` will not be migrated. Because it provides a link back to the hosting scrapbook, we will not be able to migrate this specific XPath. This will need to be a post-migration metadata update.
 
 `Example record - arrpgimg:319 <https://digital.lib.utk.edu/collections/islandora/object/arrpgimg:319/datastream/MODS/view>`_
 
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -2431,6 +2431,7 @@ I'm not sure about the relevance of `relatedItem[@type='otherVersion']/location/
 
 relatedItem/identifier[@type = 'local']
 ---------------------------------------
+
 Use Case
 ^^^^^^^^
 This XPath's attribute value (`local`) is used to indicate the manuscript number associated with the resource's archival collection.
@@ -2466,6 +2467,7 @@ Decision
 
 relatedItem/identifier[@type = 'catalog']
 -----------------------------------------
+
 Use Case
 ^^^^^^^^
 `@type='catalog'` is used exclusively in the Van Vactor collection to indicate the identifying number for an alternate version of the score.
@@ -2518,6 +2520,7 @@ Decision
 
 relatedItem/identifier[@type = 'pid']
 -------------------------------------
+
 Use Case
 ^^^^^^^^
 `@type='pid'` is used in collection-level records to indicate featured items and should not be migrated.
@@ -2551,6 +2554,7 @@ Do not migrate.
 
 relatedItem/location/url
 ------------------------
+
 Use Case
 ^^^^^^^^
 This XPath is used 8516 times, but only has 33 distinct strings.
@@ -2590,6 +2594,7 @@ The `dbo:isPartOf` property was selected.
 
 relatedItem/location/physicalLocation/
 --------------------------------------
+
 Use Case
 ^^^^^^^^
 This XPath is used once in the Charles Dabney collection. It provides an authority, a valueURI, and string value, in this single case, for the University of Tennessee's Special Collections.
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -2178,7 +2178,7 @@ relatedItem as simple structural metadata
 -----------------------------------------
 Use Case
 ^^^^^^^^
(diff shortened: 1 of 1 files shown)


docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
++-------------------------------+--------------------+-------+--------------------------------------------------------------------------------------------+
+
+relatedItem as simple structural metadata
+-----------------------------------------
+Use Case
+^^^^^^^^
+`relatedItem`, with and without attributes, is used in a number of collections to express structural relationships. These relationships will be handled/expressed by default behavior in our next-gen DAMS. In the case of `relatedItem/abstract`, the values present should be handled at the collection level.
+
+Justification
+^^^^^^^^^^^^^
```
mlhale7, line 2182: Perhaps include everything after your first sentence in the use case section above as your justification? I think it would be helpful to have some text in the justification for consistency's sake. The first sentence shows how we currently used the metadata and the rest says what we are going to do. relatedItem/abstract isn't structural metadata, so maybe just rename this category something like "elements we won't migrate" or split abstract out into something else?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+| relators:[term]               | URI/String Literal | N/A   | Use with a role from MARC Code List of Relators role terms.                                |
+|                               |                    |       | Value is either text or URI from a controlled vocabulary (like                             |
+|                               |                    |       | Library of Congress Name Authority File).                                                  |
++-------------------------------+--------------------+-------+--------------------------------------------------------------------------------------------+
+
+relatedItem as simple structural metadata
+-----------------------------------------
+Use Case
+^^^^^^^^
+`relatedItem`, with and without attributes, is used in a number of collections to express structural relationships. These relationships will be handled/expressed by default behavior in our next-gen DAMS. In the case of `relatedItem/abstract`, the values present should be handled at the collection level.
```
mlhale7: If you want, you could also include here a little bit about how these values are currently displayed or shared in the Use Case section. Currently we do have Solr fields for these values (utk_mods_relatedItem_titleInfo_title_ms) and we show the collection title in the Click for Details section using this field. Going forward we can duplicate this just using the relationship in the DAMS or just remove it as there are usually many other indicators of the larger collection on an item page.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+XPath
+^^^^^
+:code:`relatedItem[not(@*)]` OR
+:code:`relatedItem[@type='host'][@displayLabel[matches(., 'project') or matches(., 'Project') or matches(., 'Digital Collection') or matches(., 'Project Part') or matches(., 'Is Part Of')]` OR
+:code:`relatedItem/abstract`
+
+Decision
+^^^^^^^^
+Do not migrate.
+
```
mlhale7, line 2195: Even though we're not migrating, can we include some XML as an example so it's clear what information is present?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+relatedItem[@type='host'][@displayLabel='Collection']
+-----------------------------------------------------
+Use Case
+^^^^^^^^
+This XPath is used to indicate the resource's archival collection.
```
mlhale7, line 2275: If you want to expand upon the use case, you could include some more information about how these values are (or potentially might be shared). We do share this as a string in DPLA, though currently nothing is done with it. I know Gretchen had worked on a white paper on including collection names in DPLA that doesn't seem to have come to fruition on the actual platform - https://docs.google.com/document/d/16r_px4GajLIOZMlyXyN0pJFfWpstJIKxHF3brgM4YOo/editI think I might've mentioned this in a comment on your initial mapping doc.

In terms of actual current use, we actually include archival collection as a facet using utk_mods_archColl_archID_ms. It does also display in the Click for details section.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+XPath
+^^^^^
+:code:`relatedItem[@type='otherVersion']/identifier` OR
+:code:`relatedItem[@type='otherVersion']/location/url`
+
+Decision
+^^^^^^^^
+See the following section on `relatedItem/identifier[@type]`.
+
+I'm not sure about the relevance of `relatedItem[@type='otherVersion']/location/url`. In theory, it would look something like this, but it would need to be added to our metadata/RDF after migration:
```
mlhale7: I think this was added in arrowmont so that users could immediately see the context of the images (where it lies in the scrapbook/album). This text is present on the [meeting notes page for the migration](https://jirautk.atlassian.net/wiki/spaces/DLP/pages/2326636): "Search results would return individual images but there's a link under the image that would take the user to page of the scrapbook (title would also indicate what the image is a part of)". I think we need to rewrite what you have here to be more confident. I'd agree though that obviously the links themselves won't hold following a migration.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+      </location>
+    </relatedItem>
+
+.. code-block:: turtle
+
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+
+    <https://example.org/objects/1> dcterms:hasVersion <uri-of-source-scrapbook-in-new-system> .
+
+relatedItem/identifier[@type]
```
mlhale7: I think @type='local' can be grouped into the category "relatedItem[@type='series'][@displayLabel='Project']" since these are treated the same, correct? This might simplify the Use case description below (there are multiple use cases given here).

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+-----------------------------
+Use Case
+^^^^^^^^
+This XPath's `@type` has three distinct values: `local`, `catalog`, and `pid`. `@type='pid'` is used in collection-level records to indicate featured items and should not be migrated. `@type='local'` is used to indicate the manuscript number of the resource's archival collection. `@type='catalog'` is used in the Van Vactor collection to indicate the identifying number for an alternate version of the score.
+
+Justification
+^^^^^^^^^^^^^
```
mlhale7, line 2439: Some justification is needed - I don't think this is easy to do when you have three different use cases in this single category.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+        dbo:isPartOf <https://n2t.net/ark:/87290/v8pz5703> ;
+        opaque:sheetmusic_hostItem "Three songs for soprano, alto flute, English horn, and bass clarinet, M120" .
+
+`@type='pid'` will not be migrated.
+
+relatedItem/location/url
+------------------------
+Use Case
+^^^^^^^^
+This XPath is used 8516 times, but only has 33 distinct strings.
```
mlhale7, line 2559: This value is used to highlight the archival collection related to the digital collection. It aims to encourage users to further explore the archival collection by giving them a link to the finding aid.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+    </relatedItem>
+
+.. code-block:: turtle
+
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+
+    <https://example.org/objects/1> dbo:isPartOf <https://n2t.net/ark:/87290/v81g0jf1> ;
+        dbo:collection "Ruskin Cooperative Association Collection, MS.0023" .
+
+relatedItem/location/physicalLocation/
```
mlhale7, line 2594: Since this category only involves one record, would you like to write a ticket and I'll go ahead and just remove the location/physicalLocation? Let me know if you have previously and I have simply missed it! I think having a whole category for this one troublesome record adds unnecessary complexity to the mapping doc.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+        relators:cre <http://id.loc.gov/authorities/names/n81119648> ;
+        relators:cre "Jensen, Richard E." ;
+        relators:cre <http://id.loc.gov/authorities/names/no2004118058> ;
+        relators:cre <http://id.loc.gov/authorities/names/no2004118055> ;
+        relators:cre "Schock, Jack" ;
+        relators:cre <http://id.loc.gov/authorities/names/n88243079> ;
+        relators:cre <http://id.loc.gov/authorities/names/n78078895> ;
+        relators:cre <http://id.loc.gov/authorities/names/n50038965> .
+
 Hierarchical Sheet Music Identifier
```
mlhale7: I had added this to relatedItem when we were looking over identifiers as we had identified the sheetmusic_hostItem property from opaquenamespace identifier then. You have already accounted for this in the relatedItem/identifier[@type] category. Feel free to remove this from the mapping.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,6 +2152,622 @@ Drop all values in mods:part.
...
+        <title>Albert "Dutch" Roth Photograph Collection</title>
+      </titleInfo>
+    </relatedItem>
+
+.. code-block:: turtle
+
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+    @prefix opaque: <http://opaquenamespace.org/ns/> .
+
+    <https://example.org/objects/1> dbo:collection """A. G. "Dutch" and Margaret Ann Roth Papers, MS.3334""" ;
```
mlhale7: Thinking on some of the decisions we made in Mat's PR with shelfLocator, I'm not sure that we really want to retain this level of description about the archival collection. I don't think it hurts to have it, but it is possible that the Archival Series may be rearranged in the future and that this information will become inaccurate. I feel that this information doesn't really need to be duplicated in our metadata - it can live in ArchivesSpace by itself. Let me know if others have thoughts.

docs/source/contents/4_draft_mapping.rst
CanOfBees: Done!

docs/source/contents/4_draft_mapping.rst
CanOfBees: @mlhale7 - thanks for linking to the Arrowmont migration project docs. In some ways this seems this (setting up links from an item to scrapbook) should be a post-migration project, but maybe there's a compelling reason to think about it as part of our migration?

docs/source/contents/4_draft_mapping.rst
CanOfBees: I feel like a real doofus: I completely failed to read past my own work here. Apologies.

docs/source/contents/4_draft_mapping.rst
CanOfBees: As mentioned earlier, I agree with removing this section (and not migrating) - it makes our approach to handling archival metadata more consistent.

docs/source/contents/4_draft_mapping.rst
CanOfBees: Thank you for your words and thoughts - this was very helpful!

docs/source/contents/4_draft_mapping.rst
CanOfBees: @mlhale7 - I'm not sure. Since we're not migrating the `[@type='series'][@displayLabel='Project']` xpath, it doesn't seem to make sense to include a separate xpath in that section. 

I'll try splitting these into three groups.

docs/source/contents/4_draft_mapping.rst
CanOfBees: Hopefully splitting these into subsections will be helpful. As has been said repeatedly on this PR - thank you for your thoughts!

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,24 +2152,743 @@ Drop all values in mods:part.
...
+:code:`relatedItem[@type='host'][@displayLabel[matches(., 'project') or matches(., 'Project') or matches(., 'Digital Collection') or matches(., 'Project Part') or matches(., 'Is Part Of')]` OR
+:code:`relatedItem/abstract` OR
+:code:`relatedItem[@displayLabel='Featured Item']`
+
+Decision
+^^^^^^^^
+Do not migrate.
+
+`Example record - relatedItem[not(@)]: volvoices:11925 <https://digital.lib.utk.edu/collections/islandora/object/volvoices:11925/datastream/MODS/content>`_
+.. code-block:: xml
```
mathewjordan, line 2198: This is where it seems to be breaking.

docs/source/contents/4_draft_mapping.rst
mlhale7: If you can slightly edit this to remove "I'm not sure of the relevance of" to be more conclusive, I'm happy to resolve this.

mlhale7 (approved):
@CanOfBees - thanks for all of your work on this!

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2152,24 +2152,744 @@ Drop all values in mods:part.
...
+          <roleTerm authority="marcrelator" type="text" valueURI="http://id.loc.gov/vocabulary/relators/cre">Creator</roleTerm>
+        </role>
+      </name>
+    </relatedItem>
+
+.. code-block:: turtle
+
+    @prefix dbo: <http://dbpedia.org/ontology/> .
+    @prefix dcterms: <http://purl.org/dc/terms/> .
+    @prefix relators: <http://id.loc.gov/vocabulary/relators> .
```
mathewjordan: I don't recall what is correct here... **http://id.loc.gov/vocabulary/relators/** or **http://id.loc.gov/vocabulary/relators**?

mathewjordan (commented):
Other than my comment relating to relators, it looks great to me.

docs/source/contents/4_draft_mapping.rst
mathewjordan: The reason why I ask is that we (or maybe just I did) seem to do this differently across different top-level elements. I don't recall how to question and ask this using RDF inquisitor.

markpbaggett: It should be:  `http://id.loc.gov/vocabulary/relators/`

No need to use inquisitor. You're prefixing the URI so that you don't have to write everything out like you would in the old n3 way.  The old n3 way would look something like this:

```
<https://example.org/objects/1> <http://id.loc.gov/vocabulary/relatorscre> <http://id.loc.gov/authorities/names/n83189337> .
```

Look at the predicate or try to resolve in a browser.  You'll get a 404 not found.

Make sense?

markpbaggett (changes requested):
The only real feedback I have is to change that one prefix and deal with the merge conflict.

docs/source/contents/4_draft_mapping.rst
CanOfBees: If I had 5¢ for every time I've left a trailing slash off of a namespace URI, I would be able to retire and live a life of profligate excess. 

@mathewjordan thanks for catching this!

docs/source/contents/4_draft_mapping.rst
CanOfBees: Belatedly, thanks again for your help with finding this breaking point.

CanOfBees:
I merged offline to deal with the conflicts a little easier.

PR #46, opened 2020-12-03
Add overall table, tables for elements without them,  fix predicates/URIs

mlhale7:
**JIRA Issue**: [DIGITAL-925](https://jirautk.atlassian.net/browse/DIGITAL-925)

What Does this Do?
==================

This PR attempts to prepare all namespace / predicate tables for our final deliverable. A few additions have been made. Tables for the following top-level elements have been added as they weren't given in the initial mapping:

1. identifier
2. physicalDescription
3. typeOfResource
4. classification
5. recordInfo
6. accessCondition

The namespace section of the document has also been added to so that all namespaces present in the mapping document are represented.

Finally, a few smaller edits were made. For DBPedia, "dbpedia" rather than "dbo" had been used as the prefix in a few instances. In addition, some final forward slashes were added to a few of the URIs (like LoC identifiers URL - http://id.loc.gov/vocabulary/identifiers/ ). In addition, an attempt was made to standardized the predicate tables we include throughout.

How Should This Be Tested?
==========================

Answer the following:

1. Are all namespaces present in the final mapping (including relatedItem) represented in the Namespace section?
2. Do all of the top-level elements have a predicate table and are all the predicates represented?
3. Are there any instances in which a range should have been included but wasn't (currently we don't specify range for any property)?
4. Are any additional clarifications to the Usage Notes within the tables needed?
5. Is the correct value type for each predicate indicated?

Additional Notes
================

If we determine as a group that range is essentially "n/a" for all properties in our mapping, I would like to go ahead and remove that column from the tables. Please comment on this so I can take this action.


docs/source/contents/4_draft_mapping.rst
```diff
@@ -67,6 +79,20 @@ Mapping
 identifier
 ==========
 
++------------------------+------------+-------------------+---------------------------------------------------------------------------------------+
+| Predicate              | Value Type | Range (if needed) | Usage Notes                                                                           |
++------------------------+------------+-------------------+---------------------------------------------------------------------------------------+
+| dbo:isbn               | Literal    |                   | Use for identifiers with type=""isbn                                                  |
```
CanOfBees: There's a little typo here `type=""isbin`

docs/source/contents/4_draft_mapping.rst
```diff
@@ -67,6 +79,20 @@ Mapping
...
++------------------------+------------+-------------------+---------------------------------------------------------------------------------------+
+| dbo:issn               | Literal    |                   | Use for identifiers with type=""issn"                                                 |
```
CanOfBees: same; small typo `type=""issn"`

CanOfBees (changes requested):
Two small changes - otherwise looks very good.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks Bridger for your eagle eyes - mine apparently are failing me.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks for catching this. I've just committed changes for both typos and hopefully not introduced any table spacing issues in the process.

mlhale7:
All - I went back through and added a range for any properties that needed them after inspecting them in the Inquisitor. I also noticed that genre didn't have any use case column, so I added one for consistency. @CanOfBees, please feel free to suggest any necessary edits for this. Finally, some of us distinguished the header row in our tables and some didn't. I made this consistent across the document.

markpbaggett (commented):
Did we decide that we would keep range, but put a note about why we are ignoring range where appropriate?

mlhale7:
My understanding was that we'd describe our reasoning for any divergences in the justification and decision sections.

markpbaggett:
If that's the case, we should def drop that column everywhere.  The way we have range definitions now is simply not true.  We shouldn't say something is not applicable if it has a range restriction but we choose to ignore it. 

That's what I think at least.

mlhale7:
Sorry that I misunderstood from Friday. I can rework this PR and basically move out the ranges from the tables to a sentence in each decision section where it is relevant. This shows that we did consider the range even when we aren't following it. @CanOfBees and @mathewjordan - what are your thoughts?

CanOfBees:
@mlhale7 your suggestion seems appropriate to me (i.e. moving range from tables to a sentence).

markpbaggett (approved):
👍🏻

CanOfBees (approved):
I'm giving this 5 bags of popcorn :popcorn: :popcorn: :popcorn: :popcorn: :popcorn:

PR #48, opened 2020-12-08
XPaths and Stuff

CanOfBees:
**JIRA Issue**: `DIGITAL-926 <https://jirautk.atlassian.net/browse/DIGITAL-926>`_

What Does this Do?
==================
Simply: adds lots and lots and lots and lots of `:code:` prefixes and 2 3/4 hogsheads of  backticks.

Broadly: this PR 
1. removes all `mods` namespace prefixes, 
2. normalizes inline `:code:` prefixes on XPaths, 
3. wraps properties in `` `back:ticks` ``, 
4. normalizes the spelling of XPath in section headers,
5. adds a few words here and there to add clarity to a given XPath expression,
6. adds hard returns between subsections and text (or at least tries to - I probably missed one :( ),
7. and hopefully in a non-breaking-kind-of-way, made one small change to a table, on accident I swear. 

How Should This Be Tested?
==========================

Probably just a quick read, or a detailed read, if you'd prefer. I don't think we can trust tests on this one: tests will be failing on this PR, because I don't have full table modifications pulled in from @mlhale7's work on #46. I'm happy to merge those changes in (once finished), if passing tests are critical.

Additional Notes
================

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -622,7 +622,7 @@ abstract
 +------------------+------------+-------------------+-----------------------------------------------------+
 | Predicate        | Value Type | Range (if needed) | Usage notes                                         |
 +==================+============+===================+=====================================================+
-| dcterms:abstract | Literal    |                   | Use for all mods:abstracts that are not blank nodes |
+| dcterms:abstract | Literal    |                   | Use for all abstracts that are not blank nodes |
 +------------------+------------+-------------------+-----------------------------------------------------+
 
 Abstracts that are not Blank Nodes
@@ -1656,7 +1656,7 @@ note - Just a note
 Use Case
 ^^^^^^^^
 
-Usually, a note is just a note.  The xpath section below lists when this is the case. In the case that an xpath has a
+Usually, a :code:`note` is just a :code:`note`.  The XPath section below lists when this is the case. In the case that an XPath has a
 specific attribute and value, prepend the value to the text node.
 
 Justification
@@ -1668,11 +1668,15 @@ to the text node when one exists.  When one doesn't, simply take the text node.
 In Bibframe, there was no attempt to convert the 562 MARC field.  For this reason, "handwritten" documents are just
 regular notes.
 
-Xpath
+XPath
 ^^^^^
 
-`mods:note` OR `mods:note[@type="handwritten"]` OR `mods:note[@type="provenance"]` OR `mods:note[@displayLabel="Attribution"]`
-OR `mods:note[@displayLabel="use and reproduction"]` OR `mods:note[@displayLabel="Local Rights"]`
+:code:`note` OR 
+:code:`note[@type="handwritten"]` OR 
+:code:`note[@type="provenance"]` OR 
+:code:`note[@displayLabel="Attribution"]` OR 
+:code:`note[@displayLabel="use and reproduction"]` OR 
+:code:`note[@displayLabel="Local Rights"]`
 
 Decision
 ^^^^^^^^
@@ -1705,7 +1709,7 @@ note - Instrumentation
 Use Case
 ^^^^^^^^
 
-When a note has a `@type = "Instrumenation"`, it is not a general note. Instead, this element is a listing of the
+When a note has a :code:`@type = "Instrumenation"`, it is not a general note. Instead, this element is a listing of the
 performing forces called for by a particular piece of music.
 
 Justification
@@ -1716,10 +1720,10 @@ MusicBrainz, but none seemed to have a predicate to represent this idea. We did 
 Oregon Digital did have a matching predicate.  In the Samvera community, not only is this ontology used, but occasionally
 the community has suggested new predicates to be created within Opaque Namespaces.
 
-Xpath
+XPath
 ^^^^^
 
-`mods:note[@type="Instrumentation"]`
+:code:`note[@type="Instrumentation"]`
 
 Decision
 ^^^^^^^^
@@ -1747,7 +1751,7 @@ note - First Line
 Use Case
 ^^^^^^^^
 
-When a note has a `@type = "First line"` or `@type = "first line"`, it is not a general note. Instead, this element is
+When a note has a :code:`@type = "First line"` or :code:`@type = "first line"`, it is not a general note. Instead, this element is
 a direct transcription of the first line of lyrics appearing in a song.
 
 Justification
@@ -1761,7 +1765,8 @@ the community has suggested new predicates to be created within Opaque Namespace
 Xpath
 ^^^^^
 
-`mods:note[@type="First line"]` or `mods:note[@type="first line"]`
+:code:`note[@type="First line"]` OR
+:code:`note[@type="first line"]`
 
 Decision
 ^^^^^^^^
@@ -1789,7 +1794,7 @@ note - Target audience
 Use Case
 ^^^^^^^^
 
-If a note has a displayLabel attribute with the value of "Grade level", it refers to the target audience of the resource.
+If a note has a :code:`displayLabel` attribute with the value of "Grade level", it refers to the target audience of the resource.
 
 Justification
 ^^^^^^^^^^^^^
@@ -1797,10 +1802,10 @@ Justification
 The MARC 521 field should be mapped to the Bibframe intended audience field. The field is defined as information that
 identifies the specific audience or intellectual level for which the content of the resource is considered appropriate.
 
-Xpath
+XPath
 ^^^^^
 
-`mods:note[@displayLabel="Grade level"]`
+:code:`note[@displayLabel="Grade level"]`
 
 Decision
 ^^^^^^^^
@@ -1827,17 +1832,17 @@ note - Uncontrolled keyword or Tag
 Use Case
 ^^^^^^^^
 
-Some of our notes actually refer to uncontrolled keywords or tags.
+Some of our :code:`note`\ s actually refer to uncontrolled keywords or tags.
 
 Justification
 ^^^^^^^^^^^^^
 
-While not preferred, Samvera treats these as dcterms:subjects with a literal rather than an a URI.
+While not preferred, Samvera treats these as `dcterms:subjects` with a literal rather than an a URI.
 
-Xpath
+XPath
 ^^^^^
 
-`mods:note[@displayLabel="Tags"]`
+:code:`note[@displayLabel="Tags"]`
 
 Decision
 ^^^^^^^^
@@ -1862,19 +1867,22 @@ note - DPN Deposits and Other Things to Ignore
 Use Case
 ^^^^^^^^
 
-We have several notes that we do not need to migrate.
+We have several :code:`note`\ s that we do not need to migrate.
 
 Justification
 ^^^^^^^^^^^^^
 
 The data here is no longer important.
 
-Xpath
+XPath
 ^^^^^
 
-`mods:note[@displayLabel="DPN"]` OR `mods:note[text()=""]` OR `mods:note[@displayLabel="Intermediate provider"]` OR
-`mods:note[@displayLabel="Intermediate Provider"]` OR `mods:note[@displayLabel="Transcribed from Original Collection"]`
-OR `mods:note[@displayLabel="Project Part"]`
+:code:`note[@displayLabel="DPN"]` OR
+:code:`note[string()=""]` OR
+:code:`note[@displayLabel="Intermediate provider"]` OR
+:code:`note[@displayLabel="Intermediate Provider"]` OR
+:code:`note[@displayLabel="Transcribed from Original Collection"]` OR
+:code:`note[@displayLabel="Project Part"]`
 
 Decision
 ^^^^^^^^
@@ -1913,7 +1921,7 @@ Use Case
 ^^^^^^^^
 
 Several elements contain unintentional null values. There are five within Tennessee Documentary History. Additional null
-subjects include vpmoore:133 and adams:76. Most of roth seems to have null mods:subject/mods:name/mods:namePart values.
+subjects include vpmoore:133 and adams:76. Most of roth seems to have null subject/name/namePart values.
 It appears we might have inserted some blank nodes using the Islandora form entry. As there is no information, these
 "values" are not used and have no true use case.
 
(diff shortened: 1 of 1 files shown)


markpbaggett (commented):
@CanOfBees, I'm not sure I understand.  Your tests are failing because you modified those tables and they are now not formatted correctly.  I don't think I understand the plan here.  If we were to review, approve, and merge this, it will break that table.  If @mlhale7  changes the table before this is merged, there will either be a merge conflict or it will break the table when this is merged?

Why are we not making sure the table is formatted correctly even though you changed the table? I'm sure there is a reason, I just don't get it.

What's the thinking here about when this will be merged?  Are

CanOfBees:
@markpbaggett very good point - I'm not sure how I pulled spaces *out* of those tables. In any case, I see that the mistake(s) are mine. I'll get them cleaned up.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -198,12 +199,12 @@ Justification
 ^^^^^^^^^^^^^
 
 OCLC identifiers could be useful if these materials are ever shared with HathiTrust, as this value is a requirement for
-submission. Only one property, dbpedia:oclc, was identified to use and it aligns with our philosophy guidelines.
+submission. Only one property, `dbpedia:oclc`, was identified to use and it aligns with our philosophy guidelines.
```
markpbaggett: It looks like @mlhale7 has decided to go with `dbo` instead of `dbpedia`. Not sure where is best, but I'd just change this here.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -281,21 +286,25 @@ ISBNs
...
 International Standard Book Numbers are present as identifier values in the Children's Defense Fund collection. As these
 identifiers have meaning outside of the context of UTK and might be used by patrons in a search to find these materials,
 it is important that we continue to support a unique field for these values.
 
 Justification
 ^^^^^^^^^^^^^
+
 Properties for ISBN values are established in DBpedia and the Standard Identifiers Scheme. Because we give preference to
-core ontologies rather than library specific ones, we selected dbpedia:issn.
+core ontologies rather than library specific ones, we selected `dbpedia:issn`.
```
markpbaggett: I think this is `dbo` now.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1883,8 +1920,8 @@ None type
 Use Case
 ^^^^^^^^
 
-Several elements contain unintentional null values. There are five within Tennessee Documentary History. Additional null
-subjects include vpmoore:133 and adams:76. Most of roth seems to have null mods:subject/mods:name/mods:namePart values.
+Several :code:`subject` elements contain unintentional null values. There are five within Tennessee Documentary History. Additional null
+:code:`subject`\ s include vpmoore:133 and adams:76. Most of roth seems to have null :code:`subject`\ /:code:`name`\ /:code:`namePart` values.
```
markpbaggett: Curious:  why not :code:`subject/name/namePart` ?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1937,32 +1974,32 @@ Topical and name subjects with URIs
...
 
-Remediated collections include subject values with URIs.
+Remediated collections include :code:`subject` values with URIs.
 
 Justification
 ^^^^^^^^^^^^^
 
-In migration, subjects with name and topical values will be treated in the same way. We have decided that the previous
-distinction between name and topical values as subjects is not essential - only the presence of all the values in the
+In migration, :code:`subject`\ s with :code:`name` and :code:`topical` values will be treated in the same way. We have decided that the previous
```
markpbaggett: it's `topic`, not `topical`.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1937,32 +1974,32 @@ Topical and name subjects with URIs
...
+distinction between :code:`name` and :code:`topical` values as :code:`subject`\ s is not essential - only the presence of all the values in the
```
markpbaggett: It's `topic`, not `topical`.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -1937,32 +1974,32 @@ Topical and name subjects with URIs
...
+    :code:`subject[@valueURI]/name/namePart` OR
+    :code:`subject/name[@valueURI]/namePart`
 
 Decision
 ^^^^^^^^
 
-When a valueURI is present for topical or name subject, it will be the value used in migration. Examples showing each
-of the distinct Xpaths are given below:
+When a :code:`valueURI` is present for :code:`topical` or :code:`name` subject, it will be the value used in migration. Examples showing each
+of the distinct XPaths are given below:
```
markpbaggett: It's `topic`, not `topical`.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -2074,7 +2111,7 @@ Xpath
 Decision
 ^^^^^^^^
 
-String values for topical or name subjects will be migrated when a valueURI is not present.
+String values for :code:`topical` or :code:`name` subjects will be migrated when a :code:`valueURI` is not present.
```
markpbaggett: It's `topic` not `topical`.

markpbaggett (changes requested):
A couple questions and requested changes:

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks for catching this - yes, this needs to change to dbo. I changed it in the table, but not the text.

docs/source/contents/4_draft_mapping.rst
mlhale7: This is where it's hard to change text values of single elements to code. If we are going to keep with making these into code I think we do need to spell out the Xpath. It gets a bit lengthy, but it's very clear this way.

docs/source/contents/4_draft_mapping.rst
mlhale7: Yes, this was intended as text (an adjective) and not as code. If we are making these code, we'll need to change them to topic.

docs/source/contents/4_draft_mapping.rst
CanOfBees: Thanks for catching this.

docs/source/contents/4_draft_mapping.rst
CanOfBees: (-_-) I don't have any idea what I was thinking here. PEBKAC.

docs/source/contents/4_draft_mapping.rst
CanOfBees: No editing and listening to Keel. :man_shrugging: 

Thanks for catching this!

docs/source/contents/4_draft_mapping.rst
```diff
@@ -372,15 +381,15 @@ Samvera uses brackets to wrap title strings in direct mapping examples. Accordin
...
 ^^^^^
 
-`mods:titleInfo[@supplied="yes"]/mods:title`
+:code:`titleInfo[@supplied="yes"]/title`
 
 Decision
 ^^^^^^^^
 
-In these cases a supplied="yes" may also be present for one titleInfo element. Supplied titles would be used as dcterms:title. Triples will not indicate supplied titles using brackets.
+In these cases a :code:`supplied="yes"` may also be present for one :code:`titleInfo` element. Supplied titles would be used as `dcterms:title`. Triples will not indicate supplied titles using brackets.
```
mathewjordan, line 392:
```suggestion
In these cases a :code:`supplied="yes"` may also be present for one :code:`titleInfo` element. Supplied titles would be used as  :code:`dcterms:title`. Triples will not indicate supplied titles using brackets.
```

docs/source/contents/4_draft_mapping.rst
```diff
@@ -487,20 +497,22 @@ titleInfo - titleInfo has nonSort sub-element
...
 
-An object with a single titleInfo element and sub-element of nonSort.
+An object with a single :code:`titleInfo` element and sub-element of :code:`nonSort`.
 
 Justification
 ^^^^^^^^^^^^^
-We desire clean strings and will not retain nonSorts moving forward.
 
-Xpath
+We desire clean strings and will not retain :code:`nonSorts` moving forward.
```
mathewjordan:
```suggestion
We desire clean strings and will not retain :code:`nonSort` elements moving forward.
```

mathewjordan (changes requested):
@CanOfBees I scoured twice and these were the two things that popped out at me.

docs/source/contents/4_draft_mapping.rst
CanOfBees: I made a-possibly-ill-considered stylistic decision: properties will be only be backticked and XPaths will be inline-code-prefixed; i.e. `` `dbo:isPartOf` `` and `` :code:`titleInfo[\@supplied='yes']` ``.

Does that make sense? Happy to discuss alternates if you'd like.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -841,7 +855,7 @@ When there is no :code:`mods:name/@valueURI`, use the string literal from :code:
     <https://example.org/objects/1>
         relators:cre "Daniel, Charles R. (Charlie), Jr., 1930-" .
 
-If there is a :code:`mods:name/valueURI` but it's empty, use the string literal instead:
+If there is a :code:`name/valueURI` but it's empty, use the string literal instead:
```
mlhale7: I don't think we've represented this appropriately since valueURI is an attribute on name. Perhaps change it to :code: `name[@valueURI]`?

docs/source/contents/4_draft_mapping.rst
```diff
@@ -917,25 +932,25 @@ Do Not Keep Any Other Values Associated with a Name
...
-* :code:`name/role/roleTerm/@authority`
-* :code:`name/@authority`
-* :code:`name/role/roleTerm/@authorityURI`
-* :code:`name/@type`
-* :code:`name/displayForm`
-* :code:`name/description`
+:code:`name/role/roleTerm/@authority` OR
+:code:`name/@authority` OR
+:code:`name/role/roleTerm/@authorityURI` OR
+:code:`name/@type` OR
```
mlhale7, line 951: I might be missing something, but my understanding is that attributes on a particular element need to be in brackets. In other sections they are represented this way - see line 979. So here, we'd want :code:`name[@type]`, etc.). I'll go through and note each instance where this same thing occurs if we'd like to change it.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -3300,20 +3380,25 @@ Do not migrate.
...
 
 Justification
 ^^^^^^^^^^^^^
+
 We use these values for search/browse facets, as well as item-level metadata display. Additionally, collection titles are shared with DPLA.
 
 XPath
 ^^^^^
+
 :code:`relatedItem[@type='host'][@displayLabel='Collection']`
```
mlhale7, line 3397: From a consistency standpoint, it would be nice to either use single or double quotes to enclose values in attributes (e.g. @type="host" or @type='host' consistently). I think we as a group should decided. This doesn't significantly affect the actual meaning though, so I also think we can decide to not address it. @markpbaggett & @mathewjordan what are your thoughts on this?

mlhale7 (changes requested):

docs/source/contents/4_draft_mapping.rst
CanOfBees: Great catch, good question, and this is something that we should clarify:

`name/@valueURI` refers to the value of the attribute; e.g.
```xml
<name valueURI="http://www.something.net/names">Bridger</name>
```
vs `name[@valueURI]` which refers to the name element; i.e. select all `name` elements with a `valueURI` attribute. I think this particular part of our document is saying, "If there is a valueURI attibute on the name element but the attribute is empty, use the string value of the element." It should still have an `@`!!!

docs/source/contents/4_draft_mapping.rst
CanOfBees: Thanks for pointing this out; similarly to the comment on approximately line 858, these XPaths point to the attribute(s) themselves. E.g. `name[@type]` we would be saying, "Do not migrate `name` elements with a `@type` attribute!", where `name/@type` says, "Do not migrate the `@type` attribute (and its value)."

docs/source/contents/4_draft_mapping.rst
CanOfBees: It's true that I didn't think about this. Since single or double quotes are both valid in expressions (as long as they're matched or paired!), either is fine. From a consistency standpoint: I had tried (only in my work) to keep single quotes in the XPath subsections; e.g. the `originInfo` XPath section around line 1000, but then in the other subsections, used either. But I didn't apply that approach to the other XPath subsections!

If we want to make this consistent I think it would just be a matter of some time with a regex (and then some review).

docs/source/contents/4_draft_mapping.rst
mlhale7: Yes, I was not reading closely enough. We should have an @, but the rest is fine.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks for the clarification - no changes are needed. I was definitely "missing something."

docs/source/contents/4_draft_mapping.rst
mlhale7: I think in the interest of getting this PR merged, let's table this for now. I'm happy to address this in the future.

docs/source/contents/4_draft_mapping.rst
mlhale7: Thanks for adding the attribute symbol.

CanOfBees: Thank you for catching it!

docs/source/contents/4_draft_mapping.rst
CanOfBees: That sounds fine to me - and it makes sense: if I made this change, there'd need to be at least one more round of reviews.

mlhale7 (approved):
Thanks for all of the detailed work in this PR. Your changes to some syntax of the MODS elements within descriptive text addressed most of my concerns. Thanks for also adding that @.

markpbaggett (approved):
Looks good to me. 👍🏻

docs/source/contents/4_draft_mapping.rst
mathewjordan: Yeah, that makes sense. Thanks.

mathewjordan (approved):

CanOfBees:
Merged #48 into main branch.

PR #52, opened 2021-01-12
Install minimal Read the Docs theme.

mathewjordan:
What Does this Do?
==================

Installs a minimally configured Read the Docs theme for the MODS to RDF document.

How Should This Be Tested?
==========================

- Checkout my changes to the `build.yml` and the `conf.py`
- Review **build** run [#279](https://github.com/UTKcataloging/mods_to_rdf/runs/1685631869)
- I've setup a dummy doc you can see that is watching the _rtd_ branch, you can see that here: https://dumbme.readthedocs.io (I'll tear down after the merge)

Additional Notes
================
Let me know if you think we should customize this further. Things look pretty good to me right now but maybe y'all have some better ideas.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Install Sphinx
         run: sudo apt-get install python3-sphinx
+      - name: Install Read the Docs theme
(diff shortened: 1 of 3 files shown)


mathewjordan:
This is crap. I created a new one with the commits squashed.

PR #53, opened 2021-01-12
Install minimal Read the Docs theme.

mathewjordan:
What Does this Do?
==================

Installs a minimally configured Read the Docs theme for the MODS to RDF document.

How Should This Be Tested?
==========================

- Checkout my changes to the `build.yml` and the `conf.py`
- Review **build** run [1685767129](https://github.com/UTKcataloging/mods_to_rdf/runs/1685767129)
- I've setup a dummy doc you can see that is watching the _rtd-theme_ branch, you can see that here: https://dumbme.readthedocs.io (I'll tear down after the merge)

Additional Notes
================
Let me know if you think we should customize this further. Things look pretty good to me right now but maybe y'all have some better ideas.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,6 @@ jobs:
         uses: actions/checkout@v2
       - name: Install Sphinx
         run: sudo apt-get install python3-sphinx
-      - name: Install Read the Docs theme
-        run: sudo pip install sphinx-rtd-theme
       - name: Install our application with pipenv
         uses: VaultVulp/action-pipenv@v2.0.1
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Install Sphinx
         run: sudo apt-get install python3-sphinx
+      - name: Install Read the Docs theme
+        run: sudo pip install sphinx-rtd-theme
       - name: Install our application with pipenv
         uses: VaultVulp/action-pipenv@v2.0.1
         with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -12,7 +12,7 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
-          pre-build-command: "pip install pipenv && pipenv install && pipenv shell"
+          pre-build-command: "pip install pipenv && pipenv install && pipenv install sphinx-rtd-theme && pipenv shell"
           build-command: "make html"
       - name: Auto Assign Reviewers
         uses: kentaro-m/auto-assign-action@v1.1.2
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -12,7 +12,7 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
-          pre-build-command: "pip install sphinx-rtd-theme && pip install pipenv && pipenv install && pipenv shell"
+          pre-build-command: "pip install pipenv && pipenv install && pipenv shell"
           build-command: "make html"
       - name: Auto Assign Reviewers
         uses: kentaro-m/auto-assign-action@v1.1.2
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -12,7 +12,7 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
-          pre-build-command: "pip install pipenv && pipenv install && pipenv shell"
+          pre-build-command: "pip install sphinx-rtd-theme && pip install pipenv && pipenv install && pipenv shell"
           build-command: "make html"
       - name: Auto Assign Reviewers
         uses: kentaro-m/auto-assign-action@v1.1.2
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -12,7 +12,7 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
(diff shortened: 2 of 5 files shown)


mlhale7 (approved):
@mathewjordan - thanks for implementing this. I think the theme looks great. The only things I'm noticing are issues in the actual document itself, not theming. I may open another PR to address formatting on lines 222 (missing tab) and 3210/3212 (missing returns) once this is merged.

markpbaggett (changes requested):
@mathewjordan, thanks.  We're going to have to do a few things to get this working.  I'm going to try to step through this here and hopefully I won't miss anything.

First, it's important to note that our GitHub actions (build and pr) don't actually push any content  to our host.  Instead, at this point, they are instructions to install our app for testing purposes on some GitHub virtual machine.  This GitHub virtual machine is not the ReadTheDocs virtual machine. You could do this (you'll either need to write the action or find one), but that's not what's happening here right now.

At root, you'll notice there is  yaml file that tells read the docs to do on  pushes to master.  This configuration is handled in the settings for this repo.  In that yaml, you'll see a key value pair that starts with `setup_py_install`.  That value is telling read the docs to install our application by using the `setup.py` file at root.

In that file, there is a function called `setup` with an argument called `extras_require`. That argument has a dict with a list assigned to a key called `docs`.  That list needs to have this module in it (`sphinx-rtd-theme`) or readthedocs is going to fail.  It's not using the action, so it's not going to understand line 19, 49, or 85 of conf.py.  When that happens, the badge to represent Test Builds on Push will continue to Pass, but that Docs Badge is going to change to Failing. If you fix that one thing, the Actions should pass and docs should pass.

One more thing:  you installed the module in the Action, but you should also do that locally and commit your Pipfile and Pipfile.lock.  That will let other people clone this locally and do a pipenv install and be able to use this.

I may be missing something, but I think that should do it.

mathewjordan:
@markpbaggett I took a couple of steps in an attempt to make the changes you requested.

Pipenv now knows about the **sphinx-rtd-theme** package and installs from the Pipfile. I think this resolves the issue you noted.

I am able to replicate this:

1. `pipenv install && pipenv shell`
2. `cd docs && make html`
3. `open build/html/index.html` (does this type of command work in your world?)

I also attempted to only rely on this instance of the package from the pipenv install, however, you can see in the three failed commits fefbf98, 92456fb, 1bba996. In the end I needed to add the `pip install sphinx-rtd-theme` back in to get the builds to pass.

Anywho, ready for review again.

.github/workflows/build.yml
```diff
@@ -8,6 +8,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Install Sphinx
         run: sudo apt-get install python3-sphinx
+      - name: Install Read the Docs theme
+        run: sudo pip install sphinx-rtd-theme
```
markpbaggett, line 12: This isn't needed.  Pipenv will take care of this if we have the theme in our Pipfile and Pipfile.lock.

.github/workflows/pr.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
-          pre-build-command: "pip install pipenv && pipenv install && pipenv shell"
+          pre-build-command: "pip install sphinx-rtd-theme && pip install pipenv && pipenv install && pipenv shell"
```
markpbaggett, line 15: This isn't needed.  the end of the pip install will take care of installation of the package if it's in the Pipfile and Pipfile.lock.

markpbaggett (changes requested):
Looks good.  Remove those pip install instructions related to your new package and we should be good.

.github/workflows/build.yml
mathewjordan: Hey @markpbaggett I did try to remove these and yet it still fails the _build_ workflow run. See https://github.com/UTKcataloging/mods_to_rdf/pull/53/commits/fefbf98d00c355aae83b9f11f093c0cbbfb88a8e

.github/workflows/pr.yml
mathewjordan: Hey @markpbaggett I did try to remove these and yet it still fails the _build_ workflow run. See https://github.com/UTKcataloging/mods_to_rdf/pull/53/commits/1bba996d2e56277ad806bfc8cdbc74c1bcb6c156

mathewjordan:
I'm a stage of confusion on this. I did remove those lines you suggested in https://github.com/UTKcataloging/mods_to_rdf/pull/53#pullrequestreview-569956552 and the workflows fail still.

markpbaggett:
> I'm a stage of confusion on this. I did remove those lines you suggested in [#53 (review)](https://github.com/UTKcataloging/mods_to_rdf/pull/53#pullrequestreview-569956552) and the workflows fail still.

Can you post your failing action response?  I'm having trouble figuring out what is happening when.  Or, if you don't mind, can I just commit to this pull request?  I think I can work out what's happening if I do the push and read the response.

mathewjordan:
@markpbaggett Yeah, Please commit be free to commit. I'm also able to meet if it helps just to chat and watch it.

markpbaggett:
> @markpbaggett Yeah, Please commit be free to commit. I'm also able to meet if it helps just to chat and watch it.

To keep it clean I'll just checkout a new branch based on yours and work from that.

markpbaggett:
@mathewjordan, I'm at a loss.  Fundamentally, what we're doing here is wrong, but I can't get it to work differently. I tried to build multiple ways, but no matter what I do the problem persists.  There is an error that I've tried to trace down ([this is as close as I've gotten](http://www.pyinvoke.org/faq.html#i-m-getting-ioerror-inappropriate-ioctl-for-device-when-i-run-commands), but I think we just need to cheat and move on.

Apologies for giving you the run around here.  Approving.

markpbaggett (approved):
As @mathewjordan points out, this won't work if we don't do this.  Approving.

PR #54, opened 2021-01-25
Address syntax issues on lines 222, 3209, and 3212.

mlhale7:
What Does this Do?
==================

This PR address syntax errors on 222, 3209 and 3212. Code blocks were not be rendered appropriately due to these errors.

How Should This Be Tested?
==========================

Check to see if the changes address the formatting issues. Note that currently all the tests are not running appropriately due to a Sphinx issue. Comment if this problem blocks the ability to merge.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -219,7 +219,7 @@ The property `opaque:accessionNumber` was selected.
 
 .. code-block:: xml
 
-<identifier type="acquisition">1996.10.1</identifier>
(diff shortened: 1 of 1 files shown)


markpbaggett (approved):
@mlhale7 this looks good.  I just opened a pr that addresses the problem you're experiencing on builds. If you want, we can keep this open until we merge that so you can pull, but I don't think we need to do that.  I think this is fine as is since we know what is happening.

CanOfBees (approved):
@mlhale7 thanks for adding these changes - they look good to me.

mathewjordan (approved):
Approved! 🚀

PR #56, opened 2021-01-29
Top level elements links

mlhale7:
**JIRA Issue**: [DI-1374](https://jirautk.atlassian.net/browse/DI-1374)

What Does this Do?
==================

This PR adds navigation to each top-level element in the mapping.

How Should This Be Tested?
==========================

- Are all of the top-level elements present in the new contents section?
- Do the anchor links work?
- Is the "Contents" section named and placed appropriately within the doc?

Additional Notes
================

Versioning:
Do people think we should revert version 1.0 and then re add 1.0 after this is merged or do we want to go up to 1.0.1?

Merge requirements:
While everyone's feedback would be appreciated, I think we can agree to merge this as long as we have at least one approval (instead of our regular two) at this point in the project.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -77,10 +77,13 @@ Mapping Table of Contents
 *************************
 
 :ref:`identifier`
+:ref:`titleInfo`
 
 *******
 Mapping
 *******
+
+.. _identifier:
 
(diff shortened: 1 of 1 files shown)


markpbaggett (approved):
👍🏻 -- I'm going to give everyone another day or so then merge if we get no other reviews.

PR #61, opened 2021-06-08
Use CC-BY rather than a copyright statement.

markpbaggett:
What Does this Do?
==================

Uses CSS to replace the Copyright statement.

How Should This Be Tested?
==========================

Build locally with make and look at the rendered HTML in the browser.

Additional Notes
================

I tried adding the badge with JS, but no dice as this conflicts with that directly.

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -1,0 +1,14 @@
+/* replace the copyright to eliminate the copyright symbol enforced by
+   the ReadTheDocs theme but eschewed by our legal team */
+div[role=contentinfo] {
+	visibility: hidden;
+	position: relative;
+}
+
+div[role=contentinfo]:after {
+	visibility: visible;
+	position: absolute;
+	top: 0;
+	left: 0;
+	content: "Attribution 4.0 International (CC BY 4.0)";
+}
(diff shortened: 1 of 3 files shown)


mlhale7 (commented):
When I built locally I got the expected text "Attribution 4.0 International (CC BY 4.0)".  This is much more appropriate than us asserting copyright over the documentation and limiting sharing. While the badge/link would be ideal, the changes address my concerns about sharing this documentation.

One thing I noticed was that the copyright statement/license doesn't appear to be included when a pdf is generated beyond the inclusion of names previously. I couldn't generate the PDF from the page I built with the code in this PR, so I'm uncertain what appears now that the names have been replaced in the "copyright" section. If nothing about the CC BY license is present in the generated PDF, I think we should add the following to the index.rst page "This portal includes documentation about the UTK MODS to RDF Working Group. All work is made available through an `Attribution 4.0 International (CC BY 4.0) license <https://creativecommons.org/licenses/by/4.0/>`_ " That would conveniently get a functional link in and make certain that people are aware of the license even if they only have the PDF on its own.

markpbaggett:
@mlhale7 thanks.  I may be able to fix the PDF build after merge, but I have added the text to the index page.

mlhale7 (approved):

mathewjordan (approved):

PR #62, opened 2021-09-02
Add creative commons notes.

markpbaggett:
**JIRA Issue**: `EXHIBIT-94 <https://jirautk.atlassian.net/browse/EXHIBIT-94>`_

What Does this Do?
==================

We have some Creative Commons licenses that aren't valid for IIIF and are not content negotiable.  This adds a note and justification for what a CC license should look like to be both valid and content negotiable.

How Should This Be Tested?
==========================

Review the change.  Does it make sense? Are there questions.

Additional Notes
================

I haven't done an investigation of all our CC licenses but any manifest that uses a CC license that doesn't follow the pattern listed will fail IIIF validation.  Furthermore, if we are using HTTPS instead of HTTP, the license is not content negotiable and also not valid to IIIF.

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -5206,8 +5206,7 @@ This means:
 Justification
 ^^^^^^^^^^^^^
 
-DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera. Presently only the Heilman Collection
-includes a CC license.
+DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera.
 
 Creative Commons licenses should me content negotiable and valid for JSON-LD and IIIF.
 
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -5210,7 +5210,7 @@ Justification
 
 DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera.
 
-Creative Commons licenses should me content negotiable and valid for JSON-LD and IIIF.
+Creative Commons licenses should be content negotiable and valid for JSON-LD and IIIF.
 
 XPath
 ^^^^^
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -5194,11 +5194,22 @@ or one of the CC licenses. These values are used to provide users with standard 
 status of an item and how or if it can be reused. These values are currently displayed in a facet and are recommended for
 sharing with DPLA.
 
+All creative commons licenses should be content negotiable and valid.  For this to happen, this pattern must be followed:
+
+:code:`http://creativecommons.org/licenses/*/*/rdf`
+
+This means:
+
+* Use :code:`http` instead of :code:`https` as the protocol (for content negotiation and validity)
+* End in :code:`/rdf` (for content negotiation)
+
 Justification
 ^^^^^^^^^^^^^
 
-DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera. Presently only the Heilman Collection includes
-a CC license.
+DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera. Presently only the Heilman Collection
+includes a CC license.
+
+Creative Commons licenses should me content negotiable and valid for JSON-LD and IIIF.
 
 XPath
 ^^^^^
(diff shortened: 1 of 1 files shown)


mlhale7:
@markpbaggett - thanks for noticing this. This does make sense. Should I go ahead and create a ticket to update all of our CC licenses to be only include http (instead of https)? I can focus on RFTA initially in a ticket and breakdown the work by collection if that would be most helpful.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -5194,11 +5194,22 @@ or one of the CC licenses. These values are used to provide users with standard
...
+
+* Use :code:`http` instead of :code:`https` as the protocol (for content negotiation and validity)
+* End in :code:`/rdf` (for content negotiation)
+
 Justification
 ^^^^^^^^^^^^^
 
-DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera. Presently only the Heilman Collection includes
-a CC license.
+DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera. Presently only the Heilman Collection
```
mlhale7: @markpbaggett - I would suggest that we take out the sentence "Presently only the Heilman Collection includes a CC license" before merging. We now have CC licenses in the University of Tennessee Extension Special Circulars and other Project CERES collections as well as quite a few in Galston. This statement therefore is no longer true. I think we'll be using CC a lot more in the future. I'll double check all of our links and create a ticket.

mlhale7:
Final comment - could this complicate the work we did on this PR - https://github.com/DigitalLibraryofTennessee/check_and_harvest/pull/4/files? Our CC URIs will be different from that of our partner institutions and those currently listed in the check.

markpbaggett:
> Final comment - could this complicate the work we did on this PR - https://github.com/DigitalLibraryofTennessee/check_and_harvest/pull/4/files? Our CC URIs will be different from that of our partner institutions and those currently listed in the check.

I don't think so.  I think the problem right now is that we can't run validation on all our IIIF manifests because some don't meet the expectations of the validator.  That includes URIs that use HTTPS because those don't adhere to content negotiation and they also don't pass IIIF validation.

markpbaggett:
> @markpbaggett - thanks for noticing this. This does make sense. Should I go ahead and create a ticket to update all of our CC licenses to be only include http (instead of https)? I can focus on RFTA initially in a ticket and breakdown the work by collection if that would be most helpful.

I don't think it's a huge deal.  I was just doing this while I was thinking about it because I'd been using the validator so much to check complicated things and this error kept getting thrown.  I think the important thing is making a decision and trying to follow it moving forward.

mlhale7 (approved):
Thanks for removing that sentence.

markpbaggett:
@mlhale7 I just realized that public domain marks follow a different pattern since they aren't technically licenses.  I've added the pattern above.

mlhale7 (commented):
@markpbaggett - yes, thanks for making that explicit. CC0 also has a slight variation https://creativecommons.org/publicdomain/zero/1.0/ in the URL. It's also not a license, but doesn't have "mark". Perhaps we can use a "*" instead of "mark" in the public domain url example?

markpbaggett:
How likely is it that we would ever have something that is CC0? I ask because currently that pattern is not valid for IIIF.  If our goal is to stay valid according to the presentation validator.  We should open an issue and pull request.

mlhale7:
As we already have materials with CC0 for Project Ceres/Hathi Trust, it is very likely. All of the extension special circulars are examples - https://digital.lib.utk.edu/collections/islandora/object/collections%3Aagrutesc From our conversation earlier in the year with Emily, I think a goal is to get the university to apply CC0 to a lot of their materials going forward so that we can make them as open as possible.

docs/source/contents/4_draft_mapping.rst
```diff
@@ -5194,11 +5194,23 @@ or one of the CC licenses. These values are used to provide users with standard
...
+* End in :code:`/rdf` (for content negotiation)
+
 Justification
 ^^^^^^^^^^^^^
 
-DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera. Presently only the Heilman Collection includes
-a CC license.
+DPLA maps both CC licenses and Rights Statements to `edm:rights`. So does Samvera.
+
+Creative Commons licenses should me content negotiable and valid for JSON-LD and IIIF.
```
CanOfBees: Change `...should me content negotiable...` to `...should be content negotiable...`.

CanOfBees (changes requested):
@markpbaggett Sorry I didn't get the suggest feature working on this. There's a small typo, I think, in the change.

CanOfBees (approved):
@markpbaggett thanks! looks good.

PR #63, opened 2021-09-07
Remove bootstrap for required extras.

markpbaggett:
What Does this Do?
==================

Docs builds are currently failing because of an unused requirement in the build instructions.  This removes it.

How Should This Be Tested?
==========================

You can read the docs build log if you want to see why I'm doing this and why I think this will fix it.

Additional Notes
================

That's it. :shipit:


mlhale7 (approved):
I don't see any reason not to merge.

CanOfBees (approved):
@markpbaggett where's that squirrel? ... here she is :shipit:

PR #67, opened 2022-07-31
Move publication place to the correct property.

markpbaggett:
What Does this Do?
==================

Fixes small typo that points to the wrong relator that was discovered while working through migration.

How Should This Be Tested?
==========================

Do things pass? Did I guess right?

Additional Notes
================

N/A

### DIFF
--- a/docs/source/contents/4_draft_mapping.rst
+++ b/docs/source/contents/4_draft_mapping.rst
@@ -1418,7 +1418,7 @@ As part of leveraging the EDTF format, some conversion will be necessary; e.g. t
     <https://example.org/objects/1> dcterms:issued "Jun 30, 1965", "1965-06-30" ;
         dcterms:date "1964/1965" ;
         relators:pbl "University of Tennessee Theatre Department" ;
-        relators:pub <http://id.loc.gov/authorities/names/n80003889> .
+        relators:pup <http://id.loc.gov/authorities/names/n80003889> .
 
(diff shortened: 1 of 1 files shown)


docs/source/contents/4_draft_mapping.rst
```diff
@@ -1418,7 +1418,7 @@ As part of leveraging the EDTF format, some conversion will be necessary; e.g. t
     <https://example.org/objects/1> dcterms:issued "Jun 30, 1965", "1965-06-30" ;
         dcterms:date "1964/1965" ;
         relators:pbl "University of Tennessee Theatre Department" ;
-        relators:pub <http://id.loc.gov/authorities/names/n80003889> .
```
mlhale7, line 1421: Since this URI refers to the University of Tennessee, pub (publisher) is correct here rather than pup (place of publication). For it to be pup, it needs to be an actual place name rather than an organization. 

In the example in the mapping, the given pup refers to Meadville, PA (http://id.loc.gov/authorities/names/n79072935) for instance:

```
<https://example.org/objects/1> relators:pbl "Keystone View Company" ;
    relators:pup <http://id.loc.gov/authorities/names/n79072935> ;
    dcterms:created "between 1890 and 1930?", "1890?/1930" .
```

mlhale7 (commented):
I don't believe anything needs to be changed.

PR #69, opened 2022-08-19
Add arks.

markpbaggett:
What Does this Do?
==================

Adds a reference to ARKs in MODS to RDF.

How Should This Be Tested?
==========================

Does it pass validation checks.

Additional Notes
================

Nope.


mlhale7 (approved):
👍

PR #70, opened 2022-11-03
Add keyword.

markpbaggett:
What Does this Do?
==================

Splits subject into subject and keyword.

How Should This Be Tested?
==========================

Check that tests pass.


mlhale7 (approved):
@markpbaggett - thanks for this update. All changes are correct.

PR #71, opened 2022-11-05
Add coords and wgs.

markpbaggett:
What Does this Do?
==================

Adds WGS and splits coords from spatial

How Should This Be Tested?
==========================

Make sure tests pass and changes make sense.

Additional Notes
================

This maps coords to wgs -- not ica.  ICA does not have a coordinates property; it has a coordinates class.  It would only make sense to use that ontology if we were doing something like this:

``` turtle

@prefix ica: <https://www.ica.org/standards/RiC/ontology#Coordinates> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix skos: <http://www.w3.org/2004/02/skos/core#> .

<https://some-geographic-location> a ica:Coordinates ;
    rdfs:label "House Mountain" ;
    ica:latitude "35.58546" ;
    ica:longitude "-83.84444" ;
    skos:sameAs <https://sws.geonames.org/4630912> .
 

```


mlhale7 (commented):
I'm realizing looking at this that we still have both URIs and literals for place names associated with one property. I will add another property to the MAP for literal values. This PR would need to be updated so that "or Literal" is removed from the table. Let me you're okay with also including the addition of another place related property in this same PR.

markpbaggett:
sure.  i just didn't put the first time because i didn't see a local spatial element in the map doc.

mlhale7:
Definitely my fault (I did that section). I've added http://id.loc.gov/ontologies/bibframe/geographicCoverage for the property spatial_local to our MAP doc. Let me know if you have any suggestions.

mlhale7 (approved):
👍

PR #72, opened 2022-11-21
Change rps range to String and remove/replace examples that have been…

mlhale7:
What Does this Do?
==================

This PR makes it so that our M3 and the MODS to RDF documentation align. The changes make it so that a string is expected for repository values in both. This PR also updates examples that were outdated. Since the initial documentation was completed, remediation has been done to streamline some of the string values associated with UTK. For instance, "The University of Tennessee Libraries, Knoxville" and the misspelling "University of Tennesse Knoxville. Libraries" are no longer present in our metadata.

How Should This Be Tested?
==========================

Are there remaining conflicts between M3 and this documentation for repository values? Is the syntax correct?

### DIFF
--- a/docs/source/contents/4_mapping.rst
+++ b/docs/source/contents/4_mapping.rst
@@ -4448,7 +4448,7 @@ location
 +-----------------------------------+----------------+-------------------------------------------------------------------------+
 
 physicalLocation - minus UT variation
--------------------------
+-------------------------------------
 
 Use Case
 ^^^^^^^^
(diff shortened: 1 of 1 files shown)


markpbaggett (approved):
:shipit:

markpbaggett (changes requested):
This looks good, but the change can't be merged as is due to a merge conflict.  Can you address locally then we can approve and merge?

markpbaggett (approved):
Good enough for me. :shipit:

